PR #15, opened 2025-11-04
integrate scrample app

antonkri:


qnx_qemu/build/BUILD
```diff
@@ -49,8 +49,13 @@ qnx_ifs(
         "system.build",
         ":scripts",
         ":configs",
+        "@scrample//src:scrample",
```
AlexanderLanin: I'm guessing we need a separate target for a non minimal image. More like a demo image?

PR #21, opened 2025-11-07
update modules for 0.5.0 release

kgraeper:


known_good.json
```diff
@@ -2,7 +2,7 @@
   "timestamp": "2025-08-13T12:55:10Z",
   "modules": [
       {
-      "name": "score-baselibs",
+      "name": "score_baselibs",
       "hash": "5b3a469d80c6480f633ffc789ffb1277ee89fd7f",
```
evinoth1206: can this be removed?

.github/workflows/test_integration.yml
```diff
@@ -18,10 +18,28 @@ name: Bazel Build some repositories
 on:
   workflow_dispatch:
   push:
+  pull_request:
```
nicu1989: Since you added `on pull_request` trigger maybe it would be good to run only `on push` to main  branch. Now it runs twice per commit pushed if PR is opened, once for pushing to `feature_branch` and once for having a PR opened(synchronize option implicit defined).

AlexanderLanin: I'd like to postpone that post release, and get this merged while there are no conflicts

PR #35, opened 2025-11-13
Run persistency tests also in EBcLfSA

opajonk:
This adds building and running of `cpp_tests_persistency` with [the same scenario used for QNX](https://github.com/eclipse-score/reference_integration/blob/b8a1e1fff16eba97b74b44663af10c2e73668a1d/qnx_qemu/test/itf/test_persistency.py#L30). It is integrated as a separate Bazel target, similar to the `scrample` example. Hence, it is easy to understand can be removed easily if required. Drawback is the additional runtime of calling two Bazel targets. Note that it is not integrated as HI application, to avoid the additional complexity. It would not add much benefit anyway.

However, its introduction requires the usage of an updated `libacl-dev`, since the currently used one has a problem with static linking: `libacl.a` seems to have lacked `-fPIC` during its build. This, in turn, requires a slight extension of the baselibs patch: the new `libacl.a` is in a different location. Since all of this `libacl` business is a crutch anyway, this should be acceptable.

Also, the `.bazelrc` is made similar to `qnx_qemu/.bazelrc`. That caused a few more changes in the README.

### DIFF
--- a/.github/workflows/build_and_test_ebclfsa.yml
+++ b/.github/workflows/build_and_test_ebclfsa.yml
@@ -40,6 +40,6 @@ jobs:
         uses: actions/upload-artifact@v5
         with:
           name: test-logs
-          path: 
+          path: |
             ebclfsa/bazel-bin/scrample_integration/*.log
             ebclfsa/bazel-bin/persistency_integration/*.log


opajonk:
Builds are passing; however, if you feel that this poses any risk for 0.5.0, just don't merge it now - all good.

PR #54, opened 2025-12-12
Updated EB corbos Linux for Safety Applications Assets

opajonk:
With these new asses a few spurious error messages which could happen during the VM bootup / shutdown are gone.


opajonk:
> Priotr looked into those red tests today. Not sure if should be already fixed.

Please merge then when ready. Thanks :-)

FScholPer (approved):

PR #86, opened 2026-02-09
Images as single module

pawelrutkaq:
- expected that not all is aligned yet and we will follow up with improvement on main
- FIT will not work, until it is ported to ITF
- `integration` will use separate versions of modules until not fully connected with top module
- `qnx_qemu` is not removed yet but will be soon after merge
- Eb and autosd pieplines builds but may have run time checks disabled and this shall be fixed in further alignment PRs.


pawelrutkaq:
Doc will fail as there is a global issue even on main afaik.

PR #89, opened 2026-02-10
Integrate simple lifecycle examples

pawelrutkaq:


pawelrutkaq:
Okey, i checked QNX8 locally and it works. Seems today QNX does not want to give us tarballs, even after N-th rerun.

PR #104, opened 2026-02-18
Use Kyron examples from module itself

pawelrutkaq:


score_starter
```diff
@@ -14,7 +14,7 @@ mEntries = [
     ("Run Elektrobit Corbos aarch64 QEMU",
      "bazel --output_base=build/eb-aarch64 build --config eb-aarch64  //images/ebclfsa_aarch64/scrample_integration:run", "eb-aarch64"),
```
PiotrKorkus: bazel build or bazel run here?

PR #110, opened 2026-02-19
fix multiple exclusions in rust_coverage

PiotrKorkus:
each excluded target needs to be prefixed individually


pawelrutkaq:
Mergin as code quality job is passing but there are ongoing cache issue to be investigated

PR #183, opened 2026-03-12
codeql: removing the coding-standard repo

Saumya-R:
This PR adds a post-processing filter step to the SARIF recategorization workflow so that only results associated with paths under repos/ are kept in the final SARIF output.

Changes:

Runs jq after recategorization to filter SARIF runs[].results[] by artifactLocation.uri.
Overwrites the original SARIF with the filtered output.


PiotrKorkus (changes requested):
the number of warnings reduced from 84 to 1. This is incorrect.
We just want to remove `codeql-coding-standards-repo/*` from a links to particular issues

Saumya-R:
Abandoned it as have made changes in different branch.

PR #200, opened 2026-03-24
Update baselibs hash to 5d759cc

4og:


PiotrKorkus:
Done in https://github.com/eclipse-score/reference_integration/pull/224

PR #230, opened 2026-05-19
Add S-CORE v1.0 roadmap, release-gate docs, and overall-status SKILL

antonkri:
## Why

The reference integration had no planning documents or live status tracking for the v1.0 release. There was no way for contributors or stakeholders to see which modules had completed which process areas (PA1–PA5), what the release criteria for each PI were, or how to regenerate the status tables from live data without re-doing the work by hand.

## What

**Roadmap and PI planning pages** — a new `s_core_v_1/roadmap/` subtree with:
- `roadmap.rst`: scope list (12 modules), feature-completeness target, and qualifiable-state criteria (PA1–PA5 + Safety Analysis + Safety Management), structured around the S-CORE milestone timeline.
- Per-PI pages (`pi1.rst`–`pi4.rst`): concrete pass/fail criteria for each release gate (v0.8–v1.0), with cross-references into the overall-status tables so progress is always visible.
- `planning_approach.rst`: background on the PI model.

**Overall-status tracking page** — `overall_status.rst` with per-PA tables (one row per module × one column per deliverable), progress bars, and per-PA `needpie` charts driven by live sphinx-needs data from the repo.

**`overall-status` SKILL** — `.github/skills/overall-status/SKILL.md`: a 1,100-line Claude Code skill that fully specifies how to regenerate `overall_status.rst` from live `eclipse-score` GitHub repositories pinned in `known_good.json`. Covers counting rules, cell rendering, path filters, SVG chart generation, sanity checks, and multi-step procedure. Designed to be invoked by an AI agent on a nightly schedule once automation is in place; in the interim it can be run manually.

**Supporting infrastructure:**
- `docs/needs_filters.py`: custom sphinx-needs `:filter-func:` implementations for the per-PA pie charts (`std_req_status_for_area`, `area_verification_status`, `wp_tag_status`, `wp_done_automation_status`).
- `docs/conf.py`: registers `docs/` on `sys.path` (so filter functions load), monkey-patches `matplotlib` to suppress in-chart legends globally.
- `docs/_assets/custom.css`: compact table styling, progress-bar styles, and grid layout classes used by the new pages.
- `docs/index.rst`: redesigned landing page with a 3-column card grid (Status & Roadmap, Process & Tools, Code Quality, Integration, Modules, Releases).
- `docs/s_core_v_1/releases/releases.rst`: glob wildcard replaces manual per-release toctree list.
- `scripts/quality_runners.py`: auto-creates `docs/verification_report/` so CI no longer fails on first run.

## Changes

- `.github/skills/overall-status/SKILL.md` — new SKILL; covers the full data pipeline from `known_good.json` to finished RST
- `docs/needs_filters.py` — new custom sphinx-needs filter functions for per-PA needpie charts
- `docs/conf.py` — `sys.path` registration + `matplotlib` legend suppression patch
- `docs/_assets/custom.css` — new; compact table, progress-bar, and grid styles
- `docs/_assets/*.svg` — pre-generated per-PA progress figures and architecture diagrams (regenerated by the SKILL)
- `docs/index.rst` — landing page redesign with card grid
- `docs/s_core_v_1/roadmap/roadmap.rst` — new S-CORE v1.0 roadmap with module scope and qualifiable-state definition
- `docs/s_core_v_1/roadmap/overall_status.rst` — new per-PA status tables with needpie charts and progress bars
- `docs/s_core_v_1/roadmap/pi{1..4}.rst` — new per-PI release-gate pages
- `docs/s_core_v_1/roadmap/planning_approach.rst` — new PI model background
- `docs/s_core_v_1/releases/releases.rst` — glob wildcard; no more manual list maintenance
- `scripts/quality_runners.py` — auto-create `docs/verification_report/` dir to fix first-run CI failure


.github/skills/module-phase-tracker/SKILL.md
FScholPer, line 1: Can we move the skills to https://github.com/eclipse-score/.github/pull/56. It contains a copier template to centralize AI stuff

.github/skills/pi-tracker/SKILL.md
FScholPer, line 1: same here, please add it to https://github.com/eclipse-score/.github/pull/56

.github/skills/overall-status/SKILL.md
FScholPer, line 1: Please add it to https://github.com/eclipse-score/.github/pull/56

docs/needs_filters.py
FScholPer, line 1: Should be moved to https://github.com/eclipse-score/docs-as-code/blob/main/src/extensions/score_metamodel/sphinx_filters.py

FScholPer (changes requested):

docs/conf.py
```diff
@@ -13,6 +13,26 @@
 
 # Configuration file for the Sphinx documentation builder.
 
+import os as _os
+import sys as _sys
```
PiotrKorkus, line 17: why like this? why not import directly without renaming namespace?

docs/conf.py
```diff
@@ -13,6 +13,26 @@
...
+
+# Make docs/ importable so needs_filters.py (and similar helpers) can be used
+# as :filter-func: targets in sphinx-needs directives.
+_sys.path.insert(0, _os.path.dirname(_os.path.abspath(__file__)))
+
+
+def _patch_needpie_suppress_legend() -> None:
+    """Suppress all in-chart legends in sphinx-needs needpie charts."""
+    try:
+        import matplotlib.axes
```
PiotrKorkus, line 27: where matplotlib should be added as dep? In `//:ide_support` dependencies?
Or in final repo where it is used e.g. here ref_int?

docs/s_core_v_1/roadmap/pi1.rst
PiotrKorkus, line 1: Im affraid that this kind of static dashboards will be outdated very quickly and hard to sync. I dont see any script that auto generates it here

docs/s_core_v_1/releases/releases.rst
```diff
@@ -0,0 +1,21 @@
...
+   # SPDX-License-Identifier: Apache-2.0
+   # *******************************************************************************
+
+Releases
+============
+
+.. toctree::
+   :titlesonly:
+
+   S-Core v0.7 <score_07>
```
PiotrKorkus: can we put a wildcard here `*` so we dont need to maintain the list?

.github/skills/module-phase-tracker/SKILL.md
antonkri: This file doesn't exist any more

.github/skills/pi-tracker/SKILL.md
antonkri: This file doesn't exist any more.

.github/skills/overall-status/SKILL.md
antonkri: For me, .github would be the central place for SKILLs that are relevant for all repos. This one is relevant only for reference_integration repo.

docs/needs_filters.py
antonkri: Let us merge the PR, then I will move it to docs-as-code

docs/conf.py
antonkri: Fixed

docs/s_core_v_1/roadmap/pi1.rst
antonkri: Yes, true. I will take of updating it manually once  a week. Next step would be to install the agent to do it nightly. But let us first see, that it works stable.

docs/s_core_v_1/releases/releases.rst
antonkri: Done

docs/conf.py
antonkri: I think, dependency comes together with docs-as-code

docs/conf.py
```diff
@@ -13,6 +13,26 @@
...
+import sys as _sys
+
+# Make docs/ importable so needs_filters.py (and similar helpers) can be used
+# as :filter-func: targets in sphinx-needs directives.
+_sys.path.insert(0, _os.path.dirname(_os.path.abspath(__file__)))
+
+
+def _patch_needpie_suppress_legend() -> None:
+    """Suppress all in-chart legends in sphinx-needs needpie charts."""
+    try:
```
AlexanderLanin, line 26: _AI-assisted review (Claude):_

Great idea to monkey-patch the legend away globally — it keeps all the needpie charts consistent without touching each one individually. 🎉

One small concern though: `matplotlib` is imported here without being declared anywhere as a docs dependency (it's not in any `requirements*.txt`, `MODULE.bazel`, or docs-as-code bundle that I can find). The bare `except Exception: pass` means a missing `matplotlib` silently produces charts with legends again — which is probably not noticed until someone reviews the rendered output.

Could you either add `matplotlib` to the docs dependency list (wherever the docs-as-code bundle declares its Python deps), or tighten the guard so at least an `ImportWarning` is emitted in CI?

docs/s_core_v_1/roadmap/overall_status.rst
```diff
@@ -0,0 +1,846 @@
+:hide-toc:
```
AlexanderLanin, line 1: _AI-assisted review (Claude):_

Lovely work on the per-PA sections — the needpie charts and tables together make the status really easy to scan! 🙌

One navigability gap: `pi1.rst` cross-references `:ref:`overall_status_pa2``, `:ref:`overall_status_pa3``, and `:ref:`overall_status_pa5`` (lines 44–46), but there's no `.. _overall_status_pa1:` anchor for the Change Management section in this file. A future PI doc or other page that wants to link directly to PA1 would silently get a broken reference.

Would be great to add the missing `.. _overall_status_pa1:` label just above the "Process Area 1" heading, matching the pattern you've already used for PA2–PA5.

docs/s_core_v_1/roadmap/pi1.rst
```diff
@@ -0,0 +1,82 @@
+..
```
AlexanderLanin, line 1: _AI-assisted review (Claude):_

The feasibility checklist structure in the PI pages is really well thought out — it gives reviewers a clear list to work through before a release gate. 👍

This is already discussed in the existing thread (the author plans to update it manually weekly), but just to echo the concern a bit more formally: the static SVG progress figures (e.g. `pa2_impl_progress.svg`) embedded in the roadmap are committed as binary assets and won't update automatically. As a suggestion for a future step, even a short comment in the RST file pointing to the SKILL that regenerates them would help contributors know what to run. Just a forward-looking thought, not a blocker!

AlexanderLanin (commented):
_AI-assisted review (Claude):_

**Critical: 0 · Important: 1 · Suggestions: 2**

---

This is a genuinely impressive piece of work for a WIP PR! The scope here is ambitious — a full docs overhaul with a live status dashboard, per-PA needpie charts, per-PI release gate checklists, a custom filter library, and a dedicated SKILL to regenerate it all. The overall structure is very well considered and will clearly be a huge improvement to the project's visibility.

**Important**
- **`matplotlib` undeclared dependency** (`docs/conf.py`): The legend-suppression patch imports `matplotlib` at Sphinx build time, but `matplotlib` doesn't appear in any declared dependency list (requirements files, MODULE.bazel, docs-as-code bundle). The silent `except Exception: pass` means CI won't catch a missing dep — it'll just silently render charts with legends. Worth either adding it as an explicit dep or emitting at least a warning. See inline comment.

**Suggestions**
- **Missing `.. _overall_status_pa1:` anchor** (`overall_status.rst`): PA2–PA5 all have named anchors, but PA1 (Change Management) doesn't. Shouldn't cause broken refs today, but it's inconsistent and would silently break any future cross-reference. Easy one-liner fix. See inline comment.
- **SVG staleness hint** (`pi1.rst`): The committed SVG progress figures are great for initial visibility, but contributors won't know to regenerate them via the SKILL. A one-line RST comment pointing to the SKILL would go a long way. Just a forward-looking suggestion, not a blocker.

**High-value aspects worth calling out explicitly:**
- The `needs_filters.py` filter functions are clean, well-documented, and properly handle edge cases (missing IDs, empty tag sets, priority ordering). Really solid work.
- The `quality_runners.py` fix (auto-creating `docs/verification_report/` if missing) is a small but correct CI reliability improvement.
- The `releases.rst` glob wildcard is exactly the right approach — no more manual list maintenance.
- The overall landing page (`index.rst`) redesign with the grid layout is a big usability improvement over the flat toctree list.

Looking forward to the official review version of this — it's shaping up really nicely! 🎉

.github/skills/overall-status/SKILL.md
```diff
@@ -0,0 +1,1079 @@
...
+```
+
+Hard rules:
+- `.. needpie::` does **not** support `:title:` (build error).
+- Each cell needs `.. rst-class:: small-pie-cell` before its `.. needpie::`.
+- Pie chart sizing is governed by `_assets/custom.css` (class
+  `compact-overview-table`, fluid 33% per column, `!important` img override).
+  Do not add explicit pixel sizes.
+
+### 5.4 Implementation status rubric
```
AlexanderLanin: _AI-assisted review (Claude):_

**Duplicate `### 5.4` heading** — there are two sections both numbered 5.4: "Pie chart row (Process Status)" (line 574) and "Implementation status rubric" (line 623). The second should be `### 5.5`.

This matters more than a typo: an AI agent instructed to follow §5.4 will see two matching sections and can non-deterministically apply either rule or conflate them. Easy fix — renumber the second one.

.github/skills/overall-status/SKILL.md
```diff
@@ -0,0 +1,1079 @@
...
+single aggregate "Available". The cell headline is the **best** status
+across all FRs for that module (so a module with at least one Accepted FR
+still shows ✅ in summaries).
+
+```python
+# 1. List feature requests via the issue label (incl. milestone)
+def fetch_frs():
+    s = gh_raw(
+        "repos/eclipse-score/score/issues?state=all"
+        "&labels=feature_request&per_page=100 --paginate",
```
AlexanderLanin: _AI-assisted review (Claude):_

**`--paginate` is inside the URL string, not a CLI flag** — it will be sent as a literal query-parameter `%20--paginate` appended to the URL, not interpreted by `gh` as its own flag:

```python
s = gh_raw(
    "repos/eclipse-score/score/issues?state=all"
    "&labels=feature_request&per_page=100 --paginate",   # ← bug
    ...)
```

The `gh_raw` helper passes the first argument as the API path string (`gh api <path>`), so `--paginate` never reaches `gh`. With 100+ issues this silently truncates results and some modules would show `❌ Open` instead of their real FRs. It should be a separate CLI argument, passed to `subprocess.run` like:

```python
r = subprocess.run(
    ["gh", "api", "--paginate", api_path, "--jq", jq], ...)
```

(The GraphQL call in `fetch_fr_project_status` already does this correctly.)

.github/skills/overall-status/SKILL.md
```diff
@@ -0,0 +1,1079 @@
...
+| **PA3** Arch. Inspection | every `chklst_arc_inspection.rst` valid | mixed | no checklists |
+| **PA4** SW Dev Plan | `eclipse-score/score` has `docs/platform_management_plan/software_development.rst` (project-wide; same status in every row) | — | file absent |
+| **PA4** Code | source files exist outside `docs/` | — | none found |
+| **PA4** Detailed Design | `valid = total` per §3 over `dd*` directives | mixed | none |
+| **PA4** Impl. Inspection | every `chklst_impl_inspection.rst` / `chklst_dd_inspection.rst` valid | mixed | no checklists |
+| **PA5** Unit Tests | repo has `_test.cpp` / `_test.py` / `tests/` | — | none |
+| **PA5** C0/C1 Coverage | C0 = C1 = 100 % | data exists (any %) | not in `reference_integration` CI |
+| **PA5** Comp. Integration Tests | tests in module's own repo | — | none |
+e| **PA5** Static Analysis | zero-tolerance per-module CI workflow passes on `main` (clang-tidy / Clippy) | tools configured but no CI enforcement | no static-analysis config | _no link_ — Static cells render the **status only**, no source-code link. Same convention as Dynamic Analysis (§4.2). |
```
AlexanderLanin: _AI-assisted review (Claude):_

**Stray `e|` prefix on line 154** — the PA5 Static Analysis row starts with `e|` instead of `|`, which will break any Markdown parser that renders this table:

```
e| **PA5** Static Analysis | zero-tolerance ...
```

An AI agent reading this file is likely to still understand the intent, but a human reading the rendered GitHub page will see a garbled table row. Quick one-character fix 🙂

.github/skills/overall-status/SKILL.md
```diff
@@ -0,0 +1,1079 @@
...
+release-gate page) can link back via `:ref:\`overall_status_paN\``:
+
+```rst
+.. _overall_status_pa2:
+
+Process Area 2 — Requirements Engineering
+-----------------------------------------
+```
+
+Required labels: `overall_status_pa2`, `overall_status_pa3`,
```
AlexanderLanin: _AI-assisted review (Claude):_

**`overall_status_pa1` is missing from the required-labels list** — §5.0 lists the labels the agent must add before each PA heading, but only enumerates PA2–PA5:

> Required labels: `overall_status_pa2`, `overall_status_pa3`, `overall_status_pa4`, `overall_status_pa5`.

PA1 is absent. Since the agent regenerates `overall_status.rst` from scratch each run, it will consistently omit the PA1 anchor, silently breaking any cross-reference to it from future PI docs. The list should include `overall_status_pa1`.

(This is the instruction-side complement to the bug already noted in `overall_status.rst` itself — fixing one without the other means the agent will re-introduce the gap on the next regeneration.)

.github/skills/overall-status/SKILL.md
```diff
@@ -0,0 +1,1079 @@
...
+### 3.3 Inspections
+
+Each `chklst_*.rst` file contains a single `.. document::` directive whose
+indented `:status:` field is treated as the document status. Each file
+contributes `(1, 1)` if `:status: valid`, else `(0, 1)`. Files without any
+`:status:` field contribute `(0, 0)`.
+
+### 3.4 Lines of Code (PA4 Code column)
+
+LOC counts every line in source files (`.cpp .h .c .rs .py`) outside `docs/`,
```
AlexanderLanin, line 137: _AI-assisted review (Claude):_

**LOC methodology is described inconsistently between §3.4 and §6a.2** — two different measurement techniques are specified without a clear authoritative statement of which one the agent should use for the per-cell tracker table:

- §3.4 says: *"LOC counts every **line** in source files"* (i.e. line count)
- §6a.2 says: *`sum(blob.size for path in tree if is_src(path)) / 30`* (i.e. byte count ÷ 30), and §7 notes *"expect ±10% vs `cloc`"*

The §6a.2 footnote clarifies this is intentional for the **SVG charts** (byte ÷ 30 approximation). But §3.4 is the spec for the **per-module tracker table cell**, and it still says "counts every line" — which is technically not what the byte-based recipe produces.

This is worth a one-line clarification in §3.4: explicitly say whether the table cell uses the byte-÷-30 approximation (same as the charts) or a true line count, and ideally point to the `is_src` filter from §6a.2 to avoid duplication.

.github/skills/overall-status/SKILL.md
```diff
@@ -0,0 +1,1079 @@
...
+    "Config Mgmt":      "ft:config_management",
+    "Lifecycle":        "ft:health&lifecycle",
+    "Security/Crypto":  "ft:security&crypto",
+}
+KW_FALLBACK = {
+    "Baselibs":  ["baselib", "common librar", "abi compatible", "json parser"],
+    "Communication": ["ipc", "streaming", "record and replay"],
+    "Logging":   ["logging"],
+    "Persistency": ["persistency", " kvs"],
+    "Time":      [" time"],
```
AlexanderLanin, line 316: _AI-assisted review (Claude):_

**Fragile leading-space keyword for Time: `" time"`** — the Time entry in `KW_FALLBACK` uses a keyword with a deliberate leading space to avoid matching unrelated words like "runtime", "realtime", "datetime". That's a clever workaround, but it's also fragile: it won't match a title like `"Time Service for SCORE"` if the title is lowercased and the match is done from the start of the string, and it will silently miss any issue whose title starts with `Time ...` (no preceding space).

This is worth a comment in the SKILL explaining *why* the space is there, so the next person editing this doesn't "fix" it. Something like:

```python
"Time": [" time"],   # leading space to avoid matching "runtime", "datetime" etc.
```

Also worth considering whether `"feature request for time"` or `"time service"` would be safer anchors if the issue corpus is small enough to enumerate.

AlexanderLanin (commented):
_AI-assisted review (Claude):_

**SKILL.md deep-dive — Critical: 0 · Important: 2 · Suggestions: 3**

---

The SKILL is impressively thorough — it covers the full data pipeline from `known_good.json` to finished RST, with precise counting rules, manual override tables, cell layout examples, sanity-check procedures, and SVG chart generation. This is a great foundation for reproducible, AI-assisted status tracking.

**Important (will silently produce wrong output)**

- **`--paginate` inside the URL string** (line 200) — `--paginate` is embedded in the API path string rather than passed as a separate CLI flag. It gets URL-encoded and sent to the GitHub API as a query parameter, not interpreted by `gh`. With >100 FRs this silently truncates results and modules may show `❌ Open` instead of their real status. See inline comment.

- **Duplicate `### 5.4` heading** (lines 574 and 623) — two sections share the same number. An AI agent that looks up "§5.4" will see two conflicting matches. The second (Implementation status rubric) should be `### 5.5`. See inline comment.

**Suggestions**

- **`overall_status_pa1` missing from the required-labels list** (§5.0, line 413) — PA2–PA5 are enumerated but PA1 is not. Since the agent regenerates the file from scratch, it will consistently omit the PA1 anchor on every run, re-introducing the gap noted in `overall_status.rst`. See inline comment.

- **LOC methodology inconsistency** (§3.4 vs §6a.2) — §3.4 says "counts every line" while §6a.2 implements `blob.size / 30` (byte approximation). The §7 limitations note explains the approximation for the charts, but §3.4's prose still says "line count". One clarifying sentence in §3.4 pointing to the `is_src` filter and confirming the byte-÷-30 method would prevent a future agent from implementing it differently. See inline comment.

- **Fragile leading-space Time keyword** (line 283) — `" time"` with a leading space is intentional (to avoid "runtime", "datetime") but undocumented. Without the comment, the next editor will almost certainly "fix" the leading space and break Time attribution. A one-line inline comment in the SKILL would pin the intent. See inline comment.

**Separately noted:** the `e|` typo on the PA5 Static Analysis table row (line 154) is flagged inline — small but will break the rendered Markdown table.

Overall this is an excellent SKILL, and the most actionable changes are the `--paginate` bug and the duplicate section number. Everything else is forward-proofing for the next time the SKILL is edited or invoked. 🙌

docs/conf.py
AlexanderLanin: yeah from https://github.com/eclipse-score/docs-as-code/blob/590778cadc4034a62ad69dcb59b53dc51a3e5c45/src/requirements.txt#L774

docs/conf.py
AlexanderLanin: matplotlib is imported via https://github.com/eclipse-score/docs-as-code/blob/590778cadc4034a62ad69dcb59b53dc51a3e5c45/src/requirements.txt#L774

docs/s_core_v_1/roadmap/roadmap.rst
```diff
@@ -0,0 +1,202 @@
...
+   * - `NM <https://github.com/eclipse-score/inc_nm>`__
+     - Network management
+   * - `Some/IP <https://github.com/eclipse-score/someip>`__
+     - SOME/IP communication middleware
+
+Qualifiable State
+-----------------
+
+All modules follow the S-CORE process and use S-CORE tools for
+artifact generation across the following process areas: **Requirements Engineering**, **Architecture Design**, **Implementation**, **Verification**.
```
aschemmel-tech: Should add **Safety Analysis** as it is also "blue" in the "development" section.

AlexanderLanin:
_AI-assisted review (Claude):_

**Re-review after c3961d9 — Critical: 0 · Important: 3 · Suggestions: 1**

The previous findings are all addressed. Four new bugs remain; the unambiguous ones are fixed in **PR #256** (ready to merge into this branch). The others are noted below and need author input.

---

### Fixed in PR #256

1. **`comp_req_cell()` TRLC zero bug** — `trlc=True` with no directives found returned `✅ Available (0/0 comp_req [TRLC])` instead of `❌ Open`. The `trlc` flag forced a non-empty `parts` list, bypassing the zero guard. Fix: check `ct == 0 and at == 0` before building `parts`.

2. **Stale `§5.4` cross-reference in Step 6** — the heading renumber (Rollout status §5.4→§5.5) was applied to the heading but not to the `(see §5.4)` in-text reference, pointing implementers at the Pie chart row instead.

3. **`count_directives()` called but never defined** — §6a.2 told Claude to call `count_directives(...)` (total only), but only `count_directives_with_status()` exists. Replaced with `count_directives_with_status(...)[1]`.

4. **`zip()` drops rust coverage exit codes in `quality_runners.py`** — `unit_tests_summary` has one entry per module; `coverage_summary` has up to two (`_cpp`, `_rust`). `zip()` truncates to the shorter dict, silently ignoring rust coverage failures beyond the first module. Replaced with a flat check over the merged dict.

---

### Needs author input

**[Important] `NM` and `Diagnostic Services` missing from §2 repo table and Step 4 path filters**
Both modules appear in the §1 scope list but have no `known_good.json` key, own-repo name, or path filter defined anywhere in the skill. Claude cannot fetch data for them without out-of-band knowledge. If they are intentionally out of scope for now, §1 should say so; if they are in scope, §2 and Step 4 need rows for them.

**[Suggestion] PA5 Static Analysis table row has a 5th pipe-delimited column (line 160)**
The `| _no link_ — Static cells render…` text is appended as a 5th pipe after the 4-column table. Most Markdown renderers will silently drop it. The link-discipline note should either move into the 4th column or be a separate paragraph below the table.

docs/s_core_v_1/roadmap/roadmap.rst
```diff
@@ -0,0 +1,202 @@
...
+   :class-container: score-grid score-grid-compact
+   :gutter: 4
+
+   .. grid-item-card:: `Platform Management <https://eclipse-score.github.io/process_description/main/process_areas/platform_management/index.html>`__
+      :class-card: card-pa-grey
+
+      Manage the common platform, its modules and integration.
+
+   .. grid-item-card:: `Safety Management <https://eclipse-score.github.io/process_description/main/process_areas/safety_management/index.html>`__
+      :class-card: card-pa-grey
```
aschemmel-tech: @masc2023 and @aschemmel-tech suggest to have safety management also as a mandatory condition for the "Qualifiable State" - it contains the creation of the safety manuals and safety (release) package.

aschemmel-tech (changes requested):

antonkri:
Both remaining points addressed in 302ff2191a (on top of the four fixes from #256):

**[Important] NM & Diagnostic Services scope**
Clarified as PA1-only modules. §1 now has an explicit blockquote stating they have no own repository and no `known_good.json` key yet, so §2 and Step 4 deliberately have no rows for them and their PA2–PA5 cells render `❌ Open` by design. They still contribute `0` to the per-PA progress charts (§6a.2) and are excluded from the coverage weighted mean (§7), matching the existing behaviour. §2 has a cross-reference note describing what to do once their repos exist (add rows here **and** in Step 4).

**[Suggestion] PA5 Static Analysis 5th-column note**
Removed the orphaned pipe-delimited 5th column from the Static Analysis row (it was silently dropped by Markdown renderers in a 4-column table). The `no-link discipline` note is now a blockquote directly below the table and explicitly covers both Static **and** Dynamic Analysis, with the rationale (CI workflows live in the org-wide `.github` repo, not per module).

antonkri:
> _AI-assisted review (Claude):_
> 
> **Re-review after [c3961d9](https://github.com/eclipse-score/reference_integration/commit/c3961d962b267507d1363e13e8b631c90d4644f5) — Critical: 0 · Important: 3 · Suggestions: 1**
> 
> The previous findings are all addressed. Four new bugs remain; the unambiguous ones are fixed in **PR #256** (ready to merge into this branch). The others are noted below and need author input.
> 
> ### Fixed in PR #256
> 1. **`comp_req_cell()` TRLC zero bug** — `trlc=True` with no directives found returned `✅ Available (0/0 comp_req [TRLC])` instead of `❌ Open`. The `trlc` flag forced a non-empty `parts` list, bypassing the zero guard. Fix: check `ct == 0 and at == 0` before building `parts`.
> 2. **Stale `§5.4` cross-reference in Step 6** — the heading renumber (Rollout status §5.4→§5.5) was applied to the heading but not to the `(see §5.4)` in-text reference, pointing implementers at the Pie chart row instead.
> 3. **`count_directives()` called but never defined** — §6a.2 told Claude to call `count_directives(...)` (total only), but only `count_directives_with_status()` exists. Replaced with `count_directives_with_status(...)[1]`.
> 4. **`zip()` drops rust coverage exit codes in `quality_runners.py`** — `unit_tests_summary` has one entry per module; `coverage_summary` has up to two (`_cpp`, `_rust`). `zip()` truncates to the shorter dict, silently ignoring rust coverage failures beyond the first module. Replaced with a flat check over the merged dict.
> 
> ### Needs author input
> **[Important] `NM` and `Diagnostic Services` missing from §2 repo table and Step 4 path filters** Both modules appear in the §1 scope list but have no `known_good.json` key, own-repo name, or path filter defined anywhere in the skill. Claude cannot fetch data for them without out-of-band knowledge. If they are intentionally out of scope for now, §1 should say so; if they are in scope, §2 and Step 4 need rows for them.
> 
> **[Suggestion] PA5 Static Analysis table row has a 5th pipe-delimited column (line 160)** The `| _no link_ — Static cells render…` text is appended as a 5th pipe after the 4-column table. Most Markdown renderers will silently drop it. The link-discipline note should either move into the 4th column or be a separate paragraph below the table.

Done!

docs/s_core_v_1/roadmap/roadmap.rst
antonkri: Done in 706f73055a — added **Safety Analysis** to the mandatory process areas list in the Qualifiable State prose, matching the highlighted (blue) card in the Development grid.

docs/s_core_v_1/roadmap/roadmap.rst
antonkri: Done in 706f73055a — added **Safety Management** to the mandatory process areas list and switched its card in the Management grid from `card-pa-grey` to `card-pa-highlight`. Also added a short rationale paragraph noting that Safety Management produces the safety manuals and safety (release) package required for qualification.

aschemmel-tech (approved):
my findings resolved

AlexanderLanin:
_AI-assisted review (Claude):_

**Final re-review — Critical: 0 · Important: 1 · Suggestions: 0**

All previous findings resolved. One open structural thread remains before merge.

---

**[Important] Two unresolved ownership comments from @FScholPer**

- `docs/needs_filters.py` — @FScholPer asked it be moved to `eclipse-score/docs-as-code` sphinx_filters. @antonkri replied "Let us merge the PR, then I will move it". That is a reasonable pragmatic call, but it should be a tracked follow-up issue so it does not get lost — a custom `:filter-func:` module living in the integration repo is non-standard and will quietly prevent other repos from reusing the same charts.
- `.github/skills/overall-status/SKILL.md` — @FScholPer asked for consolidation into `eclipse-score/.github` (PR #56). @antonkri noted the skill is integration-specific and argued it belongs here. That is a valid position, but the thread is left open/unresolved; one of them should explicitly close it with their reasoning so reviewers can clear the conversation.

No code changes needed — just need those two threads resolved (closed or answered) so the PR conversation is clean for merge.

The code itself is in good shape: all bugs fixed, all other review rounds addressed, SKILL is correct and comprehensive. Ready to merge once the above are settled.

FScholPer (approved):

AlexanderLanin (approved):

PiotrKorkus (approved):

PR #237, opened 2026-05-21
take over release notes from score main

RolandJentschETAS:
take over release notes from score main. See also https://eclipse-score.github.io/score/main/contribute/general/folder.html#reference-integration-repository-folder-structure

The belonging PR in score is https://github.com/eclipse-score/score/pull/2921


AlexanderLanin:
Reviewed; LGTM

docs/index.rst
```diff
@@ -18,7 +18,7 @@ Reference Integration Documentation
 Newest Release Notes
 --------------------
 
-:doc:`score_releases/release_note_score_07`
+:doc:`score_releases/release_note_score_v0_7`
```
PiotrKorkus: should we remove this section if we now list all release notes below?
No need to duplicate and clutter

RolandJentschETAS: Would be fine for me. It also reduces some warnings. Not sure, if we should put the release notes section up as first point.

PiotrKorkus: might be actually a good idea

docs/conf.py
```diff
@@ -25,6 +25,14 @@
...
 ]
 
+# Exclude external release notes that are now maintained locally in docs/score_releases/
+exclude_patterns = [
+    "_collections/score_platform/docs/score_releases/score_releases/**",
+]
+
+# Suppress toc.excluded warnings from external modules referencing excluded docs
+suppress_warnings = ["toc.excluded"]
+
```
PiotrKorkus: Actually lets not introduce this workaround.

Merge score PR first and use new hash, so these needs won't be a problem anymore

PiotrKorkus (changes requested):

docs/conf.py
RolandJentschETAS: Hope, we do not end up in a deadlock

RolandJentschETAS: @PiotrKorkus : Done. After merge the score main and increase the score main version in the reference, I revert the change and it build without error.

PiotrKorkus: now it looks clean, thank you

PiotrKorkus (approved):

PR #243, opened 2026-06-01
fix broken link to s-core book in release notes

tobiwan3:
Replace link to s-core book with link to s-core handbook and update description. Currently the link in https://eclipse-score.github.io/reference_integration/main/score_releases/release_note_score_07.html#eclipse-s-core-book is broken.

### DIFF
--- a/docs/score_releases/release_note_score_v0_5_alpha.rst
+++ b/docs/score_releases/release_note_score_v0_5_alpha.rst
@@ -52,9 +52,10 @@ Highlights
 - Initial set of communication, persistency, orchestration, and base utility modules.
 - Experimental reference images for QNX, Red Hat AutoSD Linux, and EB corbos Linux for Safety Applications.
 
-Eclipse S-CORE book
--------------------
-The first version of the `Eclipse S-CORE book <https://eclipse-score.github.io/score/main/handbook/index.html>`_
+Eclipse S-CORE User’s Guide   
+---------------------------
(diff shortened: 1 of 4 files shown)


PiotrKorkus (commented):
This will happen again if someone modifies main branch again...

Can we create a need a link to it instead? So it wont be external link anymore and will always work?

tobiwan3:
Indeed the link may break again. And there are a number of links in the docs which may break. As I understand a need would create a copy of the user's guide in this documentation. Is that really intended?

PiotrKorkus:
> Indeed the link may break again. And there are a number of links in the docs which may break. As I understand a need would create a copy of the user's guide in this documentation. Is that really intended?

not a copy but here in reference_integration we do a "combo" build of documentation which collects needs from all repos and joins them in single build. So if you reference anything by need it will create a link that will work within compiled, versioned docs build. E.g. build for the release will link to specific version of need and will be true even when we change something in the future

FScholPer:
@PiotrKorkus @tobiwan3 how to continue here?

PiotrKorkus:
> @PiotrKorkus @tobiwan3 how to continue here?

resolve conficts

change http links to sphinx needs references for synced version reference

FScholPer:
@tobiwan3 please resolve the conflicts

tobiwan3:
Closed as obsolete

PR #256, opened 2026-06-08
Fix four obvious bugs from PR #230 review

AlexanderLanin:
## Why

Four unambiguous bugs were found during review of #230 that each have a single correct fix with no design ambiguity.

## What

**`comp_req_cell()` TRLC zero bug** — when called with `trlc=True` and no directives found (`ct==0, at==0`), the function returned `✅ Available (0/0 comp_req [TRLC])` instead of `❌ Open`. The `trlc` flag forced a non-empty `parts` list, bypassing the `if not parts` guard. Fix: check `ct == 0 and at == 0` unconditionally before building `parts`.

**Stale §5.4 cross-reference** — the heading renumber (Rollout status: §5.4 → §5.5) was applied to the heading but not to the in-text `(see §5.4)` in Step 6, pointing implementers at the Pie chart row instead.

**`count_directives()` called but never defined** — §6a.2 told Claude to call `count_directives(...)` (total only), but only `count_directives_with_status(...)` exists. Replaced with `count_directives_with_status(...)[1]`.

**`zip()` drops rust coverage exit codes** — `unit_tests_summary` has one entry per module, `coverage_summary` has up to two (`_cpp`, `_rust`). `zip()` truncates to the shorter dict, silently ignoring rust coverage failures beyond the first module. Replaced with a flat check over the merged dict.

## Changes

- `.github/skills/overall-status/SKILL.md` — fix `comp_req_cell()` zero guard, update §5.4→§5.5 cross-ref, fix `count_directives` reference
- `scripts/quality_runners.py` — replace `zip()`-based exit-code check with a flat check over both summary dicts

### DIFF
--- a/.github/skills/overall-status/SKILL.md
+++ b/.github/skills/overall-status/SKILL.md
@@ -807,13 +807,13 @@ def comp_req_cell(contents, *, trlc=False):
     for c in contents:
         v, t = count_directives_with_status(c, ('comp_req',)); cv+=v; ct+=t
         v, t = count_directives_with_status(c, ('aou_req',));  av+=v; at+=t
+    if ct == 0 and at == 0: return "❌ Open"
     parts = []
     if ct or trlc: parts.append(f"{cv}/{ct} comp_req" + (" [TRLC]" if trlc else ""))
     if at:         parts.append(f"{av}/{at} AoU")
-    if not parts:  return "❌ Open"
     body = " + ".join(parts)
     tot_v, tot_t = cv + av, ct + at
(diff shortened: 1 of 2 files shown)


antonkri (approved):

PiotrKorkus:
@AlexanderLanin please fix formatting and we can merge

AlexanderLanin:
Of course there is formatting 🤣

PiotrKorkus (approved):
We dont need to wait for all CI jobs, as main branch is NOT the target of this PR

PR #257, opened 2026-06-08
Fix duplicate releases listing and add data collection date

antonkri:
## Summary

Two small documentation fixes:

1. **Releases page duplication** — `docs/s_core_v_1/releases/index.rst` was being picked up by the glob toctree in `releases.rst`, which caused all release notes to be listed twice (once under the *Release Notes* group and once individually). Removed the redundant `index.rst`; `releases.rst` already collects all release notes via `:glob: *`.

2. **Data collection date on Overall Status** — Added a highlighted `important` admonition at the top of `docs/s_core_v_1/roadmap/overall_status.rst` so readers can see when the snapshot was taken.

## Verification

- Build the docs locally and check that the Releases page lists each release note only once.
- Confirm the *Data collected on* banner is visible at the top of the Overall Status page.


docs/s_core_v_1/roadmap/overall_status.rst
```diff
@@ -3,6 +3,10 @@
 Overall Status
 ==============
 
+.. important::
+
+   **Data collected on: 2026-06-02**
+
```
PiotrKorkus, line 9: Is `/overal-status` skill updating this field without explicit instruction to do it?

PiotrKorkus (approved):

Issue #229, opened 2026-05-19
ref_int - add showcases


PiotrKorkus:
### What

Improve current `/showcases` implementation by extending with more examples for modules that are integrated (based on `known_good.json`) and are missing their showcases.

Extend standalone showcases with the examples directly from Modules - same as it is currently done for score_communication.

Create additional complex showcases that will combine multiple Modules.

### Acceptance Criteria (DoD)

Additional examples have been added.
Showcases are building and running for all available configurations.

PR #262, opened 2026-06-15
Add logging showcase

ShivamPatidar24:
## Summary

This PR adds a new Logging example showcase demonstrating local and remote logging.

The example generates logs for all supported severity levels.

## Related

Issue: #229

## Verification

- [x] DEBUG, INFO, WARN, ERROR and FATAL logs generated
- [x] Logs visible in console output
- [x] Logs visible in DLT Viewer and Chipmunk

## Documentation

Added README and screenshots.


rmaddikery (changes requested):
Thanks for the contribution. 
If the goal is to showcase, then I would recommend you to kindly re-use the existing logging tests instead of creating new ones. This will avoid any duplication and maintenance burden.

PR #270, opened 2026-06-24
Saumya feature lifecycle cross module integration 

Saumya-R:
Cross-module integration tests verify that lifecycle management integrates correctly with other S-CORE modules in end-to-end scenarios:

- **Lifecycle ↔ Persistency** — Data persistence across recovery actions
- **Lifecycle ↔ State Manager** — Control interface and run-target management


feature_integration_tests/test_cases/tests/lifecycle/test_lifecycle_persistency_recovery.py
```diff
@@ -0,0 +1,622 @@
...
+        # Determine which supervised app is available based on version
+        supervised_app_name = f"{version}_supervised_app"
+
+        # Ensure daemon is running
+        assert daemon.is_running(), "Launch Manager daemon not running"
+
+        # Step 1: Locate the supervised process managed by the daemon.
+        # The fixture configures the daemon with the version-specific supervised app,
+        # so this must succeed.
+        supervised_pid = _find_supervised_process(daemon, supervised_app_name)
```
PiotrKorkus, line 470: **C1 — CRITICAL: the persistency writer and the killed process are different processes, so this test is meaningless.** In `test_persistency_recovery_with_daemon_supervision`, KVS data is written by `_run_persistency_probe` (lines 475, 529) — an *independent* scenario subprocess writing to `work_dir/"kvs_supervised"` (line 455). The process force-killed (line 488) is the daemon-supervised `{version}_supervised_app` (found here at line 467), which the daemon launches with only `IDENTIFIER`/`PROCESSIDENTIFIER`/`CONFIG_PATH` env vars (`daemon_helpers.py:699-708`) — it never touches `kvs_supervised`. So the 'data intact after crash' checks (lines 521, 544) are trivially true: an unrelated process's already-flushed data cannot be affected by the crash. The test would pass even if recovery corrupted the supervised app's real state or did nothing at all. It verifies no recovery↔persistency interaction. **Fix:** have the supervised app itself own the KVS storage that is killed and recovered, and assert on *its* data.

feature_integration_tests/test_cases/tests/lifecycle/test_lifecycle_state_manager_if.py
```diff
@@ -0,0 +1,192 @@
...
+    """
+    Validate the control-interface path for activate_target and status queries.
+
+    The daemon configuration contains run targets (Startup and fallback),
+    so the initial activation already exercises the activate_target path.
+    Where available, a direct IPC call is sent to trigger an additional
+    run-target switch, and log evidence is checked.
+    """
+
+    def test_status_query_returns_current_run_target(
```
PiotrKorkus: **C2 — CRITICAL: this test never sends a request; it only sleeps and greps boot logs.** `test_status_query_returns_current_run_target` does `time.sleep(1.0)` → `is_running()` → `get_logs()` → substring-scan. No status query and no `activate_target` request is ever issued. The control interface could be entirely broken (socket never opened, query endpoint missing) and the test would still pass from unrelated boot logging.

feature_integration_tests/test_cases/tests/lifecycle/test_lifecycle_state_manager_if.py
```diff
@@ -0,0 +1,192 @@
...
+# ── Explicit control-interface IPC errors ─────────────────────────────────────
+_CONTROL_ERROR_SIGNALS = [
+    "control interface: socket error",
+    "Failed to bind control socket",
+    "IPC connection refused",
+    "activate_target: error",
+]
+
+
+def _try_send_activate_target(
```
PiotrKorkus, line 70: **C3 — CRITICAL: `activate_target` verification is dead code.** `_try_send_activate_target` (this function) is never called, and `_ACTIVATE_TARGET_TOKENS` (line 50) is never used (grep-confirmed). So the activate_target / run-target-switch path this class advertises is entirely inert — `request_run_target_start` / `switch_run_targets` / `control_commands` could regress with no test signal.

feature_integration_tests/test_cases/tests/lifecycle/test_lifecycle_state_manager_if.py
```diff
@@ -0,0 +1,192 @@
...
+    "RunTarget",
+    "switch_run_target",
+    # Fallback: any run-target name the daemon logs during a transition
+    "startup",
+    "running",
+    "fallback",
+]
+
+# ── Positive evidence of a status query being answered ────────────────────────
+_STATUS_QUERY_TOKENS = [
```
PiotrKorkus: **C4 — CRITICAL: self-fulfilling assertion — status tokens match any normal boot log.** `_STATUS_QUERY_TOKENS` includes generic substrings `"status"`, `"State"`, `"Startup"`, plus run-target names `"startup"`/`"running"`/`"fallback"` (the comments even admit these are transition names, not query responses). Any daemon that logs a startup transition satisfies the assertion at line 186 regardless of whether a query was answered — effectively `assert daemon logged something`.

feature_integration_tests/test_cases/tests/lifecycle/test_lifecycle_persistency_recovery.py
```diff
@@ -0,0 +1,622 @@
...
+    return None
+
+
+@pytest.mark.daemon
+@add_test_properties(
+    partially_verifies=[
+        "feat_req__lifecycle__process_failure_react",
+        "feat_req__lifecycle__monitor_abnormal_term",
+        "feat_req__persistency__store_data",
+        "feat_req__lifecycle__restart_on_failure",
```
PiotrKorkus: **C5 — CRITICAL: traceability to non-existent requirement IDs (+ semantically wrong trace).** `feat_req__lifecycle__restart_on_failure` and `feat_req__lifecycle__recovery_action` (lines 301-302) **do not exist** in the requirements doc (verified). The valid IDs are `recovery_action_support` and `recov_run_target_switch`. The trace is also semantically wrong: the test's own docstring (lines 442-443, 486) states the recovery action is `switch_run_target` → `fallback_run_target`, **not** a restart — so `restart_on_failure` is doubly incorrect. Fix the IDs and use `recov_run_target_switch`.

feature_integration_tests/test_cases/tests/lifecycle/test_lifecycle_persistency_recovery.py
```diff
@@ -0,0 +1,622 @@
...
+            pytest.skip(
+                f"Scenario binary {target} not found in runfiles. "
+                "This test requires the scenario binary to be available as a data dependency."
+            )
+    else:
+        if target not in _binary_path_cache:
+            _binary_path_cache[target] = build_tools.find_target_path(target)
+        scenario_binary = _binary_path_cache[target]
+
+    # NOTE: C++ all_value_types scenario requires the full FitScenario fixture infrastructure
```
PiotrKorkus: **C6 — CRITICAL: C++ and Rust run *different* persistency scenarios; a real C++ defect is papered over.** `_run_persistency_probe`/`_simulate_probe_crash` use `persistency.default_values.checksum` for **cpp** but `persistency.supported_datatypes.all_value_types` for **rust** (lines 101-108, 206-211). The comment here admits the C++ `all_value_types` scenario errors with 'Invalid value type' and switches to checksum 'as a workaround.' So (a) the two languages validate different persistency behavior — a C++-specific regression in `all_value_types` is never caught, and (b) a genuine C++ scenario/CLI defect is hidden instead of reported. Both scenarios exist and write identical data in both languages, so the workaround is not due to a missing scenario.

feature_integration_tests/test_cases/tests/lifecycle/test_lifecycle_persistency_recovery.py
```diff
@@ -0,0 +1,622 @@
...
+        # After SIGKILL the LCM detects unexpected termination within milliseconds and
+        # immediately transitions to fallback_run_target. We poll the log for evidence of
+        # this transition; the full shutdown sequence runs in the background and can take
+        # up to ~30 s (multiple supervision + transition cycles).
+        print("Waiting for LCM to detect crash and initiate recovery...")
+        recovery_deadline = time.time() + 10.0
+        recovery_detected = False
+        while time.time() < recovery_deadline:
+            logs = daemon.get_logs()
+            if "unexpected termination" in logs.lower() and "fallback" in logs.lower():
```
PiotrKorkus: **C7 — CRITICAL: recovery detection is a loose log-grep; 'fallback' is present regardless of any crash.** Detection requires `"unexpected termination" in logs AND "fallback" in logs` (this line). But `fallback_run_target` is part of the daemon topology and is logged at startup (before any kill), so the `"fallback"` clause is satisfied unconditionally — the check degrades to just `"unexpected termination"`, which can be emitted by normal supervision churn of other components. Recovery could be broken and the test still pass. The 10 s deadline / 0.2 s poll (lines 497, 504) is also timing-fragile under load.

feature_integration_tests/test_cases/tests/lifecycle/test_lifecycle_state_manager_if.py
```diff
@@ -0,0 +1,192 @@
...
+            return False
+
+    # No control tool available — the initial run-target activation performed
+    # by the daemon itself on startup already exercises the activate_target path.
+    return False
+
+
+@pytest.mark.daemon
+@add_test_properties(
+    partially_verifies=[
```
PiotrKorkus, line 120: **C8 — CRITICAL: requirement over-claiming — 7 requirements, one loose log-grep.** This `partially_verifies` claims `logic_arc_int__lifecycle__controlif`, `request_run_target_start`, `switch_run_targets`, `start_named_run_target`, `run_target_support`, `control_commands`, `query_commands`. The single test exercises none of activation/switching/named-start/query — it only checks a generic token exists in boot logs. Six of seven requirements are marked verified without exercised behavior → false coverage. Trim `partially_verifies`.

feature_integration_tests/test_cases/daemon_helpers.py
```diff
@@ -0,0 +1,754 @@
...
+    input_json.write_text(json.dumps(dynamic_config, indent=2))
+
+    json_out_dir = work_dir / "json_out"
+    flatbuffer_out_dir = work_dir / "flatbuffer_out"
+    json_out_dir.mkdir(exist_ok=True)
+    flatbuffer_out_dir.mkdir(exist_ok=True)
+
+    # When running inside a Bazel test sandbox, `bazel build/info` commands fail
+    # because the sandbox has no MODULE.bazel workspace file.
+    _under_bazel = bool(os.environ.get("TEST_SRCDIR") or os.environ.get("RUNFILES_DIR"))
```
PiotrKorkus, line 307: **C9 — CRITICAL: under Bazel/CI the dynamic daemon config is discarded; no supervised app is launched.** `copy_dynamic_flatbuffer_daemon_configs` builds a dynamic config from `input_config`, but when `TEST_SRCDIR`/`RUNFILES_DIR` is set it takes the `_under_bazel` branch (this line), copies the **pre-built** flatbuffers (compiled from the *static* template whose `initial_run_target "Startup"` has `depends_on: []` and lists the supervised apps only under a never-initial `SupervisedApps` target), and returns — throwing the dynamic config away. Result: the daemon launches **no** supervised process, so `_find_supervised_process` returns `None` and the hard assertion at `test_lifecycle_persistency_recovery.py:468` fails (not xfail/skip). The recovery tests cannot pass under Bazel until the Bazel branch compiles the dynamic config (or the template's Startup depends on the supervised app + a sandbox uid/gid). *(Needs confirmation in the actual CI invocation.)*

feature_integration_tests/test_cases/tests/lifecycle/test_lifecycle_persistency_recovery.py
```diff
@@ -0,0 +1,622 @@
...
+    """
+
+    @pytest.fixture(scope="class")
+    def build_tools(self, request: pytest.FixtureRequest, version: str) -> BuildTools:
+        """Provide BuildTools instance for locating scenario binaries."""
+        from testing_utils import BazelTools
+
+        return BazelTools(option_prefix=version)
+
+    def test_persistency_continuity_across_recovery(
```
PiotrKorkus: **M1 — MED: `test_persistency_continuity_across_recovery` performs no recovery (misnamed/misleading).** The name and docstring claim it verifies snapshot stability 'during lifecycle recovery,' but it takes no daemon fixture, kills nothing, and triggers no recovery — it just runs the probe twice. It exercises none of the lifecycle requirements listed on the class; it is a 'two sequential writers to one dir' persistency test.

feature_integration_tests/test_cases/tests/lifecycle/test_lifecycle_persistency_recovery.py
```diff
@@ -0,0 +1,622 @@
...
+        proc.kill()
+        proc.wait()
+
+    assert proc.returncode == -signal.SIGKILL, (
+        f"Expected SIGKILL exit (returncode {-signal.SIGKILL}), got {proc.returncode}. "
+        "Crash simulation did not exercise abnormal termination."
+    )
+
+
+def _wait_for_process_restart(daemon: Any, process_name: str, old_pid: int, timeout_s: float = 10.0) -> int | None:
```
PiotrKorkus: **M2 — MED: docstrings/name say the app is 'restarted'; the actual action is switch-to-fallback (+ dead restart helper).** The module docstring (lines 20-25) and class docstring (line 312) describe recovery as a process **restart**, contradicting the switch_run_target-to-empty-fallback reality. `_wait_for_process_restart` (this function, lines 255-292) is defined but never called — leftover from an abandoned restart design that reinforces the misleading docs.

feature_integration_tests/test_cases/tests/lifecycle/test_lifecycle_persistency_recovery.py
```diff
@@ -0,0 +1,622 @@
...
+            "kvs_parameters": {
+                "instance_id": 1,
+                "dir": str(kvs_dir),
+            },
+        },
+    }
+    config_json = json.dumps(config)
+    command = [str(scenario_binary), "--name", scenario_name, "--input", config_json]
+
+    proc = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
PiotrKorkus, line 276: **M3 — MED: `_simulate_probe_crash` SIGSTOPs before KVS is opened, so 'crash while storage open' may be false.** `SIGSTOP` is sent immediately after this `Popen` (lines 236-247) with no synchronization that the scenario has opened/written KVS. On a slow exec the child is frozen during startup and killed before touching storage, so the documented 'abnormal termination of a process that has the KVS storage open' (claim at lines 590-591) is not actually exercised.

feature_integration_tests/test_cases/tests/lifecycle/test_lifecycle_persistency_recovery.py
```diff
@@ -0,0 +1,622 @@
...
+        "kvs_parameters_1": {
+            "kvs_parameters": {
+                "instance_id": 1,
+                "dir": str(kvs_dir),
+            },
+        },
+    }
+    config_json = json.dumps(config)
+
+    # Try both argument formats
```
PiotrKorkus: **M4 — MED: 'try both argument formats' re-runs the whole scenario on any non-zero exit.** If variant 1 (`--name/--input`) runs the scenario but exits non-zero for a legitimate reason (scenario assertion, partial write), variant 2 (`-n/-i`) executes the entire scenario **again** — masking the real failure whenever the second form succeeds and double-writing storage, which perturbs the mtime/ID diffing logic used elsewhere (lines 406-414, 534-537).

feature_integration_tests/test_cases/tests/lifecycle/test_lifecycle_persistency_recovery.py
```diff
@@ -0,0 +1,622 @@
...
+
+        # Step 1: Write initial persistency snapshot using proper infrastructure
+        _run_persistency_probe(build_tools, version, kvs_dir, timeout_s=30.0)
+
+        # Step 2: Verify snapshot was created and hash is correct
+        snapshot_files = list(kvs_dir.glob("kvs_1_*.json"))
+        assert len(snapshot_files) > 0, "Initial persistency snapshot was not created"
+
+        # Read and verify snapshot integrity
+        snapshot_data = read_kvs_snapshot(kvs_dir, instance_id=1, snapshot_id=0)
```
PiotrKorkus: **M5 — MED: hardcoded `snapshot_id=0` assumption.** Multiple reads/verifies (lines 375, 478, 521, 544, 585, 605) assume the current snapshot is always `kvs_1_0.json`, even though the first test itself computes variable snapshot IDs (lines 383-392). If the KVS rotates/increments IDs, `read_kvs_snapshot(..., snapshot_id=0)` either raises `FileNotFoundError` or reads a stale snapshot 0 while current data lives in a higher-numbered file — another self-fulfilling pass verifying the wrong file.

feature_integration_tests/test_cases/daemon_helpers.py
```diff
@@ -0,0 +1,754 @@
...
+        if self.process is not None:
+            raise RuntimeError("Daemon already started")
+
+        # Create log file
+        self.log_file = self.working_dir / "launch_manager.log"
+        self._log_fd = open(self.log_file, "w", encoding="utf-8")
+
+        # Start daemon process
+        cmd = [str(self.daemon_binary), str(self.config_file)]
+        self.process = subprocess.Popen(
```
PiotrKorkus, line 487: **M6 — MED: supervised child processes leaked on daemon stop (no process group / `start_new_session`).** `stop()` SIGTERMs then `kill()`s only the daemon PID (lines 492-503); this `Popen` has no `start_new_session=True` and there is no `os.killpg`. If the daemon is SIGKILLed or fails to propagate shutdown, supervised children are orphaned and keep running, polluting the next class's `psutil` child scan / IPC identities → flaky cross-class failures.

feature_integration_tests/test_cases/daemon_helpers.py
```diff
@@ -0,0 +1,754 @@
...
+        # If a stale process reference exists from a previous run, clear it.
+        if self.process is not None and self.process.poll() is not None:
+            self.process = None
+
+        if self.process is not None:
+            raise RuntimeError("Daemon already started")
+
+        # Create log file
+        self.log_file = self.working_dir / "launch_manager.log"
+        self._log_fd = open(self.log_file, "w", encoding="utf-8")
```
PiotrKorkus: **M7 — MED: log file truncated on every `start()`.** `open(self.log_file, "w")` truncates the constant log path on each start. For a class-scoped daemon that is stopped/restarted mid-class, a later `get_logs()` assertion (e.g. `test_lifecycle_state_manager_if.py:174`) loses pre-restart messages. Use append mode or a per-start filename.

feature_integration_tests/test_cases/tests/lifecycle/test_lifecycle_state_manager_if.py
```diff
@@ -0,0 +1,192 @@
...
+        time.sleep(1.0)
+
+        assert daemon.is_running(), "[activate_target] Daemon not running; cannot validate status-query path."
+
+        logs = daemon.get_logs()
+
+        ctrl_errors = [s for s in _CONTROL_ERROR_SIGNALS if s in logs]
+        assert not ctrl_errors, f"[activate_target] Control-interface IPC errors on status-query path: {ctrl_errors}"
+
+        if "Operation not permitted" in logs and ("setuid(" in logs or "setgid(" in logs):
```
PiotrKorkus: **M8 — MED: `pytest.xfail` guard can mask real failures.** If logs contain `"Operation not permitted"` and `setuid(`/`setgid(`, the test xfails (lines 179-184) before the real assertion. In any sandboxed CI lacking uid/gid privileges this branch is taken every run, so the test never asserts and silently becomes a non-signal even if the daemon is genuinely broken.

feature_integration_tests/test_cases/daemon_helpers.py
```diff
@@ -0,0 +1,754 @@
...
+        self.process = subprocess.Popen(
+            cmd,
+            cwd=self.working_dir,
+            stdout=self._log_fd,
+            stderr=subprocess.STDOUT,
+            text=True,
+        )
+
+        # Wait for daemon to initialize
+        time.sleep(startup_timeout)
```
PiotrKorkus, line 497: **M9 — MED: fixed sleeps instead of readiness polling (flaky).** `start()` sleeps a hardcoded 2.0 s (this line) then a single `poll()`; `is_running()` only confirms the daemon PID, never that the supervised app launched/registered. The state-manager test's `time.sleep(1.0)` (`test_lifecycle_state_manager_if.py:170`) is the only sync before reading logs. On slow/loaded CI the supervised child/logs may not exist yet → intermittent failures. Poll for the condition instead.

feature_integration_tests/test_cases/tests/lifecycle/test_lifecycle_state_manager_if.py
```diff
@@ -0,0 +1,192 @@
...
+
+        Pass/fail
+        ---------
+        PASS  At least one status/state token in daemon logs.
+        FAIL  Explicit IPC error OR no status token found.
+        """
+        daemon = launch_manager_daemon["daemon"]
+        time.sleep(1.0)
+
+        assert daemon.is_running(), "[activate_target] Daemon not running; cannot validate status-query path."
```
PiotrKorkus: **M10 — MED: mislabeled assertion messages ('[activate_target]') on the status-query test.** All three failure messages (lines 172, 177, 187) are prefixed `"[activate_target]"` though the test is about status query and never sends activate_target — copy-paste that will misdirect triage.

feature_integration_tests/itf/test_lifecycle.py
```diff
@@ -0,0 +1,12 @@
...
+#
+# See the NOTICE file(s) distributed with this work for additional
+# information regarding copyright ownership.
+#
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
```
PiotrKorkus: **L1 — LOW: `itf/test_lifecycle.py` contains only a license header (no test, no wiring).** The branch adds only the 12-line Eclipse header — no imports, no tests. If it was meant to register the new lifecycle tests into the itf suite, that wiring is missing; otherwise remove the empty placeholder.

feature_integration_tests/test_cases/lifecycle_scenario.py
```diff
@@ -0,0 +1,298 @@
...
+from collections.abc import Generator
+from pathlib import Path
+from typing import Any
+
+import pytest
+from fit_scenario import FitScenario, temp_dir_common
+from testing_utils import BazelTools
+
+
+def read_launch_manager_config(config_path: Path) -> dict[str, Any]:
```
PiotrKorkus: **L2 — LOW: dead helpers in `lifecycle_scenario.py` (and root-only defaults).** `read_launch_manager_config`, `create_launch_manager_config`, `create_daemon_integrated_config`, `add_supervised_component`, `copy_test_app_to_daemon_workspace` (lines 31, 48, 107, 191, 238) are referenced nowhere (grep-confirmed). Besides being dead, the config builders hardcode `uid/gid: 0` and would fail in unprivileged CI if ever wired up.

feature_integration_tests/test_scenarios/cpp/src/scenarios/lifecycle/launch_manager_support.cpp
```diff
@@ -0,0 +1,894 @@
...
+                            }
+                        }
+                    }
+                }
+            }
+        }
+
+        std::cout << "Testing process security configuration" << std::endl;
+        std::cout << "Process UID: " << uid << ", GID: " << gid << std::endl;
+        std::cout << "Supplementary groups: [100, 200]" << std::endl;
```
PiotrKorkus, line 380: **L3 — LOW: 17 print-only lifecycle scenarios are added but exercised by no test on this branch, and they diverge C++/Rust.** The `launch_manager_support` scenarios (process_security, process_resources, process_arguments, conditional_launching, run_targets, dependency_ordering, parallel_launching, …) are compiled/registered in both languages but **no Python test added on this branch invokes them**. Where they would be used they are not equivalent across languages, e.g.: C++ hardcodes `Supplementary groups: [100, 200]` (this line) and `CPU affinity: [0, 1]` (`.cpp:439`) while Rust reads them from input (`.rs:293-299,331-334`); Rust builds a real `HealthMonitor` in dependency/parallel scenarios while C++ only prints (`.rs:104-122,164-182` vs `.cpp:155-170,188-234`); default/'switching' strings and missing-field handling differ (`.cpp:335-336,499-507,601-615` vs `.rs:265-268,364-376,434-443`); and `LifecycleTestInput` is strict in Rust but defaulted in C++ (`.rs:28-32,98-100` vs `.cpp:65,79`). Either wire tests to these scenarios (and make the two implementations equivalent) or drop them.

feature_integration_tests/test_cases/daemon_helpers.py
```diff
@@ -0,0 +1,754 @@
...
+                    )
+
+                for candidate in candidates:
+                    if candidate.exists() and candidate.is_file():
+                        return candidate
+
+    return None
+
+
+def get_binary_path(target: str, version: str = "rust") -> Path:
```
PiotrKorkus: **L4 — LOW: `get_binary_path` `version` parameter is unused/misleading.** `version` is declared and documented but never used (path resolution keys only on `target` and `FIT_BAZEL_CONFIG`). The per-language differentiation actually comes from the version-specific `target` the caller passes; the unused param invites the false assumption that the daemon binary differs per language.

.github/workflows/test_and_docs.yml
```diff
@@ -73,6 +73,7 @@ jobs:
       - name: Execute Feature Integration Tests
         run: |
           bazel test --config=linux-x86_64 //feature_integration_tests/test_cases:fit
+          bazel test --config=linux-x86_64 //feature_integration_tests/test_cases:fit_daemon
```
PiotrKorkus: dont modify workflow fle, handle by addidng tests to existing suite

PiotrKorkus (changes requested):

Saumya-R:
Almost all review comments have been fixed. Closing this PR to split it into smaller reviewable chunks.

Issue #264, opened 2026-06-17
ref_int - update test workflow according to DR8


PiotrKorkus:
### What

According to https://eclipse-score.github.io/score/main/design_decisions/DR-008-int.html test_and_docs workflow needs to be updated to reflect described process in option 4.

Rough overview:
1. resolve deps by bazel
2. in each module (inside bazel directories or checkout?) patch deps to match resolved list
3. execute UTs, coverage etc.
4. collect in upstream as before

### Acceptance Criteria (DoD)

Testing workflow matches Option 4

LittleHuba:
PR #282 highlights a completely broken testing strategy in the reference_integration.
By testing modules as dependencies instead of root modules you force the modules to spill all their guts (e.g. make all development targets and their dependendies public API).

This leads to tight coupling and makes the integration of a module extremely hard and error-prone.
Please fix this ASAP, to enable a meaningful integration of modules.

PR #278, opened 2026-07-03
Add resolved-dependency override mechanism for DR-008

Subramanian-K812:
## DR-008: resolved-dependency resolve + override mechanism (PR 1 of 2)

Part of #264. This is the first of two PRs that split the original DR-008 Option 4 change. This PR
adds the mechanism; the follow-up PR wires it into the test workflow.

### What this PR does

Adds `ResolvedDependencies` — the mechanism DR-008 Option 4 needs to (1) resolve the full integrated
dependency set at the reference_integration root and (2) override those resolved versions directly
into an individual module.

- **Resolve (root):** merges the override directives ref_int actually declares (`MODULE.bazel` +
  `bazel_common/*.MODULE.bazel`) with the post-MVS registry versions from
  `bazel mod graph --output=json`, and serializes the result to a single `resolved_versions.json`
  manifest (`--export` / `from_mod_graph` / `to_file`). 156 modules today.
- **Override (into a module):** scans a module's `MODULE.bazel` for every declared `bazel_dep` and
  appends a `git_override` / `single_version_override` for each one we have a resolved version for,
  so the module builds/tests against the resolved set (`scan` / `overwrite`). Injection is
  append-only and idempotent (marked with begin/end comments), always skips the module-under-test
  itself (the root is never overridden), always overwrites any dependency the module already
  declares an override for (ref_int's resolved version always wins), and is never committed back to
  module sources. A declared dependency with no entry in the resolved set is left to resolve on its
  own and logs a warning.
- **Report:** the export also writes a `resolved_pins_report.json` sidecar recording, per pin,
  whether ref_int explicitly asserted it or only inherited it from MVS, whether any consumer in
  ref_int's own graph wanted a different version, and any override ref_int declares that could not be
  carried into the manifest at all (today: `rules_boost`'s `archive_override`). Nothing here changes
  what gets pinned or fails the export — it makes the existing resolve step's decisions visible.

While testing this against a real checkout it turned up that `rules_oci` was silently falling out of
the manifest entirely (its `git_override` pinned a mutable tag, which can't become a manifest entry,
and nothing warned about it) — fixed by pinning it to a commit instead.

### Files

| File | Change |
|---|---|
| `scripts/known_good/resolved_dependencies.py` | New — the `ResolvedDependencies` mechanism (resolve + override + report) + CLI |
| `scripts/known_good/tests/test_resolved_dependencies.py` | New — 67 self-contained unit tests (no cloned repos / Bazel needed) |
| `scripts/known_good/BUILD` | New — `py_library :known_good`, `py_binary :resolve_deps` (Bazel-runnable), and `score_py_pytest :known_good_tests` so the unit tests run under Bazel |
| `.github/workflows/internal_tests.yml` | Modified — runs `:known_good_tests` in CI |
| `bazel_common/score_images.MODULE.bazel` | Modified — `rules_oci` pinned by commit instead of tag |
| `bazel_common/score_test_artifact_versions.MODULE.bazel` | New — asserts the versions of the three deps Stage 2 collects test artifacts from (`googletest`, `rules_rust`, `score_toolchains_rust`); no build change today |
| `MODULE.bazel` | Modified — `include()`s the file above |
| `MODULE.bazel.lock` | Modified — mechanical consequence of the pin above |

### Follow-up (PR 2)

The two-stage `test_and_docs.yml` (Stage 1 resolves + exports the manifest; Stage 2 checks out each
module, overrides deps via this mechanism, and runs its UT + coverage; then aggregates), the
module-context `quality_runners.py`, the dependency pin bumps, and the associated config all land in
the follow-up PR, which is stacked on this one.


scripts/tooling/BUILD
```diff
@@ -81,6 +81,13 @@ py_binary(
     visibility = ["//visibility:public"],
 )
 
+# Alias: expose resolve_deps under //scripts/tooling for `bazel run //scripts/tooling:resolve_deps`.
+alias(
+    name = "resolve_deps",
```
a-zw, line 86: Why this alias? This is usually for backwards-compatibility but since this is new functionality it should not be necessary.

scripts/known_good/tests/test_resolved_dependencies.py
```diff
@@ -0,0 +1,287 @@
...
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+"""Unit tests for ResolvedDependencies (DR-008 Option 4 dependency injection).
+
+Self-contained: builds the resolved set from a temporary known_good.json and
+overwrites a temporary module MODULE.bazel — no cloned repos or Bazel required.
```
a-zw, line 16: I like this self-contained aspect. It allows to use this in an hermetic way.

However, the tradeoff is that we cannot parallelize the followup actions (build, test, docs, etc) via GitHub actions, can we?

Subramanian-K812: No, they can still run in parallel. Stage 1 resolves once and uploads resolved_versions.json as an artifact; each Stage 2 module job downloads that same artifact and does its own local override + build/test independently, still as a normal parallel matrix. The hermetic is only about the unit tests not needing real clones/bazel to validate scan()/overwrite() — it doesn't force any job serialization in the actual pipeline.

scripts/tooling/BUILD
Subramanian-K812: The alias under //scripts/tooling provides reachability alongside other entry points. The implementation itself remains in //scripts/known_good since that’s where its models and tests live. The pipeline continues to call //scripts/known_good:resolve_deps directly; the alias simply exposes it without duplicating the target.

scripts/known_good/BUILD
```diff
@@ -0,0 +1,48 @@
...
+score_py_pytest(
+    name = "known_good_tests",
+    srcs = glob(["tests/**/*.py"]),
+    data = ["//:known_good.json"],
+    pytest_config = "//:pyproject.toml",
+    deps = [":known_good"],
+)
+
+# Runnable binary for the resolve + inject workflow.
+#   Stage 1 (export): bazel run //scripts/known_good:resolve_deps -- \
```
OliverHeilwagen: Please extend description for Stage 1.
`bazel mod graph --output json > graph.json` is required as prerequisite.

In the script please make use of `BUILD_WORKING_DIRECTORY` or `BUILD_WORKSPACE_DIRECTORY` environment variable so the graph.json file can be found without requiring to be in `data = ["..."]`

scripts/known_good/BUILD
```diff
@@ -0,0 +1,48 @@
...
+#       --mod-graph graph.json --export artifacts/resolved_versions.json
+#   Stage 2 (inject): bazel run //scripts/known_good:resolve_deps -- \
```
OliverHeilwagen: Please keep Stage 1 and Stage 2 in sync.
`--resolved-deps artifacts/` or `_resolved_deps` in both.

scripts/known_good/resolved_dependencies.py
```diff
@@ -0,0 +1,510 @@
...
+from __future__ import annotations
+
+import argparse
+import json
+import logging
+import re
+import sys
+from pathlib import Path
+
+_HERE = Path(__file__).resolve().parent
```
OliverHeilwagen, line 37: Please make use of `BUILD_WORKING_DIRECTORY` or `BUILD_WORKSPACE_DIRECTORY` environment variable instead.

scripts/known_good/resolved_dependencies.py
```diff
@@ -0,0 +1,510 @@
...
+            if name and commit and remote:
+                modules.append(Module(name=name, hash=commit, repo=remote))
+
+        for match in _SINGLE_VERSION_BLOCK_RE.finditer(text):
+            body = match.group("body")
+            name = _field(body, "module_name")
+            version = _field(body, "version")
+            if name and version:
+                modules.append(Module(name=name, hash="", repo="", version=version))
+
```
OliverHeilwagen, line 453: Please add support for [multiple_version_override ](https://bazel.build/rules/lib/globals/module#multiple_version_override).

scripts/known_good/resolved_dependencies.py
```diff
@@ -0,0 +1,510 @@
...
+            # Drop comment-only lines first: hand-written MODULE.bazel files contain
+            # commented-out overrides (e.g. "# git_override(... rules_rpm ...)") that must
+            # not be captured. Inline trailing comments (after a value) are left intact.
+            text = "\n".join(ln for ln in Path(f).read_text().splitlines() if not ln.lstrip().startswith("#"))
+            for module in cls._parse_override_file(text):  # git_override + single_version_override
+                resolved[module.name] = module
+            for m in re.finditer(r'(archive_override|local_path_override)\(\s*module_name\s*=\s*"([^"]+)"', text):
+                unrepresentable.append(f"{m.group(2)} ({m.group(1)})")
+
+        graph = json.loads(Path(mod_graph_json).read_text())
```
OliverHeilwagen, line 382: Please check file existence before using it.

scripts/known_good/resolved_dependencies.py
```diff
@@ -0,0 +1,510 @@
...
+
+    # Export mode (Stage 1): build the manifest by merging the override directives ref_int
+    # declares (root MODULE.bazel + bazel_common/*.MODULE.bazel) with the resolved registry
+    # versions from 'bazel mod graph'.
+    if args.export is not None:
+        if args.mod_graph is None:
+            raise SystemExit("--export requires --mod-graph (output of 'bazel mod graph --output=json')")
+        repo_root = _HERE.parents[1]
+        override_files = [repo_root / "MODULE.bazel", *sorted((repo_root / "bazel_common").glob("*.MODULE.bazel"))]
+        override_files = [f for f in override_files if f.is_file()]
```
OliverHeilwagen:
```suggestion
        override_files = [
            f for f in [repo_root / "MODULE.bazel", *sorted((repo_root / "bazel_common").glob("*.MODULE.bazel"))]
            if f.is_file()
        ]
```

scripts/known_good/resolved_dependencies.py
```diff
@@ -0,0 +1,510 @@
...
+        return
+
+    # Inject mode (Stage 2): overwrite a module's MODULE.bazel with the resolved set.
+    if args.module_bazel is None:
+        raise SystemExit("module_bazel is required unless --export is given")
+
+    if args.resolved_deps:
+        resolved = ResolvedDependencies.from_resolved_artifact(args.resolved_deps)
+    else:
+        resolved = ResolvedDependencies.from_known_good(args.known_good_path)
```
OliverHeilwagen: Transitive dependency pinning gap

`overwrite()` only injects overrides for modules the module-under-test **directly declares**
via `bazel_dep`. Transitive dependencies silently fall through to MVS resolution, which may
pick a different version than what ref_int validated against.

**Example:**

```
ref_int (resolved: flatbuffers @ 25.12.19)
    └── communication (module under test)
            └── bazel_dep(name = "score_baselibs")   ← override injected ✅
                    └── flatbuffers            ← transitive, NO override injected ❌
                                                 MVS picks its own version
```

**Fix:** For every module in the resolved set that is _not_ directly declared by the
module-under-test, inject both a `bazel_dep` stub and the override into the injection block:

```python
# brings the transitive dep into the root graph so the override is valid
bazel_dep(name = "flatbuffers", version = "0.0.0")
git_override(module_name = "flatbuffers", commit = "...", remote = "...")
```

This also means `from_known_good` is insufficient for the inject mode, as it only carries
first-party score modules and has no transitive registry versions; `--resolved-deps` (from the
Stage-1 manifest) must be the required source for Stage 2 injection.

In Stage-1 we should also takeover and store the given `graph.json` as it is needed for Stage-2 to be aware of the necessary transitive overrides for a given MODULE.bazel file. Identification can be done via regex for `module(name = "...")`.

OliverHeilwagen (changes requested):
Please fix transitive dependency pinning gap

scripts/known_good/BUILD
OliverHeilwagen: fixed please resolve comment.

scripts/known_good/BUILD
OliverHeilwagen: fixed please resolve comment.

scripts/known_good/resolved_dependencies.py
OliverHeilwagen: Please use repo_root  in case mod_graph is not a absolute path e.g.;
```
if not mod_graph.is_absolute():
    mod_graph = repo_root / mod_graph
```

Please apply the same pattern to `export` otherwise the files are only written to runfiles dir.

scripts/known_good/resolved_dependencies.py
OliverHeilwagen: fixed, please resolve comment.

scripts/known_good/resolved_dependencies.py
OliverHeilwagen: fixed, please resolve comment.

scripts/known_good/resolved_dependencies.py
OliverHeilwagen: fixed, please resolve comment

scripts/known_good/resolved_dependencies.py
```diff
@@ -0,0 +1,1058 @@
...
+    # declares (root MODULE.bazel + bazel_common/*.MODULE.bazel) with the resolved registry
+    # versions from 'bazel mod graph'.
+    if args.export is not None:
+        if args.mod_graph is None:
+            raise SystemExit("--export requires --mod-graph (output of 'bazel mod graph --output=json')")
+        mod_graph = Path(args.mod_graph)
+        if not mod_graph.is_file():
+            raise SystemExit(
+                f"--mod-graph {mod_graph} does not exist. Produce it first with: "
+                "bazel mod graph --output=json > graph.json"
```
OliverHeilwagen:
```suggestion
                "bazel mod graph --verbose --output=json > graph.json"
```
verbose version is required.

scripts/known_good/resolved_dependencies.py
Subramanian-K812: Fixed.

scripts/known_good/resolved_dependencies.py
```diff
@@ -0,0 +1,1070 @@
...
+          forced to the exact version ref_int resolved (MVS is graph-global, so a module's own
+          subgraph could otherwise select a different version).
+
+        Produce the graph with ``--verbose`` to populate ``originalVersion``, without which every
+        report verdict is ``unknown``. The export succeeds either way.
+        """
+        resolved: dict[str, Module] = {}
+        provenance: dict[str, str] = {}
+        declared_overrides: dict[str, dict[str, str]] = {}
+        declared_deps: dict[str, dict[str, object]] = {}
```
OliverHeilwagen: It would be nice to have more descriptive data classes or enum classes to improve the readability.
May also be taken up as a follow up cleanup.

e.g.: 
`provenance`:
```
class PinProvenance(StrEnum):
    ASSERTED = "asserted"
    INCIDENTAL = "incidental"

provenance: dict[str, PinProvenance] = {}
```

`declared_overrides`:
```
@dataclass(frozen=True)
class UncarriedOverride:
    module_name: str
    kind: str
    source_file: str
    reason: str
    consequence: str

    def to_dict(self) -> dict[str, str]:
        return {
            "module": self.module_name,
            "kind": self.kind,
            "file": self.source_file,
            "reason": self.reason,
            "consequence": self.consequence,
        }
declared_overrides: dict[str, UncarriedOverride] = {}
```

`declared_deps`:
```
@dataclass(frozen=True)
class DependencySpec:
    module_name: str
    version: str | None
    dev_dependency: bool
    source_file: str

declared_deps: dict[str, DependencySpec] = {}
```

scripts/known_good/resolved_dependencies.py
```diff
@@ -0,0 +1,1070 @@
...
+                        "consequence": _UNCARRIED_CONSEQUENCE,
+                    },
+                )
+                continue
+            resolved[name] = Module(name=name, hash="", repo="", version=version)
+            provenance[name] = "incidental"
+
+        # The completeness guard: every override ref_int declared that did not become a manifest entry.
+        uncarried = [entry for name, entry in sorted(declared_overrides.items()) if name not in resolved]
+        report = _build_report(resolved, provenance, uncarried, declared_by, _internal_drift(resolved, declared_deps))
```
OliverHeilwagen: as mentioned above report would also benefit from a dataclass approach:
```
@dataclass(frozen=True)
class PinReport:
    pin: dict[str, str]
    provenance: PinProvenance
    declared_versions: list[str]
    verdict: str
    direction: str | None

    def to_dict(self) -> dict[str, object]:
        return {
            "pin": self.pin,
            "provenance": self.provenance.value,
            "declared_versions": self.declared_versions,
            "verdict": self.verdict,
            "direction": self.direction,
        }
```

PiotrKorkus (approved):

PiotrKorkus:
Hi @OliverHeilwagen, there was a discussion in todays SCORE alignment call and we decided to enable this workflow in parallel to the old integration tests. This will allow us to verify and tackle all the issues with the workflow itself in the transition period as `pull_request_target` does not allow to run it directly in the PR. On top of that, the changes are big and started to get messy - it will be easier to fix and review in separate PRs.

I will let myself merge this PR to enable rebase in #280 and adding this DR8 as parallel workflow. Please do not hesitate to continue reviewing in the part 2 PR or creating the issues directly. Thank you for contributing to the review!

OliverHeilwagen:
@PiotrKorkus Thanks for the info, i will bring my pending comments over to #280 and continue the review there.

PR #280, opened 2026-07-07
DR-008 Option 4: two-stage test-execution workflow (PR 2 of 2)

### LINKED ISSUE #264: ref_int - update test workflow according to DR8

Subramanian-K812:
# DR-008 Option 4: two-stage test-execution workflow (PR 2 of 2)

Closes #264. This is the **second** of two PRs. It is **stacked on PR 1**
(`Subramanian-K812_resolve_override_mechanism`, "resolved-dependency resolve + override
mechanism") — please review/merge that one first. Until PR 1 merges, this PR's base branch
is the PR-1 branch, so the diff shown here is only the test-execution delta.

## What this PR does

Rewires the quality workflow to the DR-008 Option 4 flow, using the resolve + override
mechanism added in PR 1:

1. **Stage 1 — integration:** builds the platform, runs Feature Integration Tests, and
   exports the resolved dependency set as the `stage1-resolved-deps` artifact
   (`resolved_versions.json`).
2. **prepare_matrix:** derives the Stage-2 module list dynamically from
   `known_good.json` (`target_sw`) — never hardcoded.
3. **Stage 2 — per module:** checks the module out at its `known_good` commit, overrides
   its `MODULE.bazel` with the Stage-1 resolved set (PR 1's mechanism), and runs the
   module's own unit tests + coverage as the Bazel root (`//...`).
4. **aggregate:** consolidates Stage 1 + Stage 2 into one quality report (and the
   release-tag test-report ZIP).

## Files

**Workflow & runners**
- `.github/workflows/test_and_docs.yml` — two-stage restructure (+ parallel docs build)
- `scripts/quality_runners.py` — module-context mode (`--module-dir` / `--resolved-deps`); calls `ResolvedDependencies.overwrite`
- `scripts/aggregate_quality_report.py` — **new**, consolidated report
- `scripts/known_good/list_modules.py` — **new**, dynamic Stage-2 matrix from `known_good.json`
- `scripts/integration_test.py` — derive build targets from `known_good.json` (drop `build_config.json`)

**Resolved pins + their config ripple (kept together so the tree is self-consistent)**
- `known_good.json` — pin bumps (baselibs, lifecycle) + per-module `bazel_config` + lifecycle `code_root_path`/`extra_test_config`
- `bazel_common/score_basic_bazel.MODULE.bazel` — flatbuffers bump
- `bazel_common/score_modules_target_sw.MODULE.bazel` — regenerated (baselibs, lifecycle)
- `MODULE.bazel.lock` — updated lock
- `rust_coverage/BUILD` — lifecycle query `//src/...` → `//score/...` (layout moved by the bump)
- `showcases/simple_lifecycle/BUILD` — lifecycle bin path `//src/...` → `//score/launch_manager`
- `.bazelrc` — coverage atomic-gcov / `-no-coverage`; eb-aarch64 outline-atomics link fix

## Known Stage 2 integration findings (surfaced, not fixed here)

The workflow surfaces two module-owned integration failures caused by the
baselibs bump (`score_logging` `CheckSizeValid` stale `uint8_t` assumption;
`score_persistency` `to_string` on `std::string_view`).


ci/stage2/CONSTRAINTS.md
```diff
@@ -0,0 +1,221 @@
+# Stage 2 — Constraint Sheet
+
+Living document. Amend it when a constraint turns out to be wrong, and record why in the
```
PiotrKorkus: If thats AI generated and maintained, where is a skill / agent for that purpose?

bazel_common/score_test_artifact_versions.MODULE.bazel
PiotrKorkus, line 1: where is this file generated from? All modules should be from known_good

bazel_common/score_rust_toolchains.MODULE.bazel
```diff
@@ -12,7 +12,10 @@
 # *******************************************************************************
 bazel_dep(name = "rules_rust", version = "0.68.1-score")
 
-bazel_dep(name = "score_toolchains_rust", version = "0.8.0", dev_dependency = True)
+# 0.9.1 declares the four *_rules_rust_miri toolchains score_lifecycle_health and
+# score_persistency register in their own .bazelrc; 0.8.0 has none. ref_int cannot un-register
+# them, so the version must match here. Locked in score_test_artifact_versions.MODULE.bazel.
```
PiotrKorkus: remove this comment

.github/workflows/test_and_docs.yml
```diff
@@ -67,41 +83,62 @@ jobs:
...
+        uses: bazel-contrib/setup-bazel@0.18.0
+        with:
+          bazelisk-cache: true
+          disk-cache: ${{ github.workflow }}-docs
+          repository-cache: true
+          cache-save: ${{ github.event_name == 'push' }}
+      - name: Set up Python 3
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
```
PiotrKorkus: why needed?

scripts/BUILD
```diff
@@ -0,0 +1,45 @@
...
+score_py_pytest(
+    name = "quality_scripts_tests",
+    srcs = glob(["tests/**/*.py"]),
+    pytest_config = "//:pyproject.toml",
+    deps = [":quality_scripts"],
+)
+
+# One label for every Python unit test, so CI runs all of them by naming a single target.
+test_suite(
+    name = "python_tests",
```
PiotrKorkus: change name, its meaningless without looking into target

.github/workflows/test_and_docs.yml
```diff
@@ -22,21 +32,31 @@ on:
   push:
     branches:
       - main
+      - Subramanian-K812_test_execution_upgrade # TEMP: remove after CI verified
```
PiotrKorkus: it doesnt work because the same workflow name is with `pull_request_target` on main

.github/workflows/test_and_docs.yml
```diff
@@ -128,10 +165,204 @@ jobs:
...
+    name: "Stage 2 — Module UT & Coverage (${{ matrix.module.name }})"
+    needs: [stage1_integration, prepare_matrix]
+    strategy:
+      fail-fast: false
+      matrix:
+        module: ${{ fromJSON(needs.prepare_matrix.outputs.modules) }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clean disk space
+        uses: eclipse-score/more-disk-space@v1
```
PiotrKorkus: use 1.1 version

.github/workflows/test_and_docs.yml
```diff
@@ -128,10 +165,204 @@ jobs:
...
+      # Prove it from the module's own post-MVS graph rather than assuming the injection took.
+      - name: Verify module resolved to ref_int's dependency versions
+        if: always()
+        run: |
+          # The resolution gate already captured this graph, before the tests ran and under the
+          # resolution they were pinned to. Reuse it rather than recomputing a second one.
+          if [ ! -s _module/module_graph.json ]; then
+            echo "::warning::no module graph captured for ${{ matrix.module.name }}"; exit 0
+          fi
+          PYTHONPATH=scripts python3 scripts/known_good/verify_stage2_resolution.py \
```
PiotrKorkus: if script needs a PYTHONPATH to be set its very bad design. Fix with proper modules imports

known_good.json
```diff
@@ -3,68 +3,75 @@
...
                         "//score/json/examples:*",
-                        "//score/flatbuffers:version_reader_test"
+                        "//score/flatbuffers/examples/config_usecase:demo_app"
                     ],
                     "langs": [
                         "cpp"
+                    ],
+                    "bazel_config": [
+                        "stage2-gcc",
+                        "stage2-rust"
```
PiotrKorkus: it feels like it doesnt need to be placed in this json as its common for every module

known_good.json
```diff
@@ -3,68 +3,75 @@
...
                         "//score/mw/com/impl:runtime_test",
                         "//score/mw/com/impl/configuration:config_parser_test",
                         "//score/mw/com/impl/configuration:configuration_test",
-                        "//score/mw/com/impl/configuration:configuration_json_parsing_strategy_test",
                         "//score/mw/com/impl/tracing/configuration:tracing_filter_config_parser_test",
                         "//score/mw/com/impl/tracing:tracing_runtime_test",
                         "//score/mw/com/impl/bindings/lola/tracing:tracing_runtime_test"
+                    ],
+                    "bazel_config": [
+                        "stage2-rust"
```
PiotrKorkus: why com has only rust?

PiotrKorkus (changes requested):
as the workflow has been completely reworked please create a private fork and execute it there, so we can see the proof of working in CI

pull_request_target wont allow to execute it here

.github/workflows/test_and_docs.yml
Subramanian-K812: Bumped to v1.1

.github/workflows/test_and_docs.yml
Subramanian-K812: Removed.

scripts/BUILD
Subramanian-K812: Renamed to all_python_unit_tests.

.github/workflows/test_and_docs.yml
Subramanian-K812: Removed.

bazel_common/score_test_artifact_versions.MODULE.bazel
Subramanian-K812: not generated — known_good.json pins first-party repos by commit, these are registry deps pinned by version (confirmed via bazel mod graph --verbose: 2 of the 3 pins do move MVS's selection)

ci/stage2/CONSTRAINTS.md
Subramanian-K812: Removed.

bazel_common/score_rust_toolchains.MODULE.bazel
Subramanian-K812: Removed.

known_good.json
Subramanian-K812: score_communication passes use_base_constraints_only=True to its own gcc.toolchain(), which changes its generated toolchain target name leading to referencing a target that doesn't exist.

known_good.json
Subramanian-K812: score_time and score_communication are exceptions for the above.

.github/workflows/test_and_docs.yml
Subramanian-K812: Fixed.

Subramanian-K812:
Ran the reworked workflow end-to-end on a private fork:
https://github.com/Subramanian-K812/reference_integration-ci-verify/actions/runs/32220961400

scripts/known_good/resolved_dependencies.py
```diff
@@ -0,0 +1,1064 @@
...
+          forced to the exact version ref_int resolved (MVS is graph-global, so a module's own
+          subgraph could otherwise select a different version).
+
+        Produce the graph with ``--verbose`` to populate ``originalVersion``, without which every
+        report verdict is ``unknown``. The export succeeds either way.
+        """
+        resolved: dict[str, Module] = {}
+        provenance: dict[str, str] = {}
+        declared_overrides: dict[str, dict[str, str]] = {}
+        declared_deps: dict[str, dict[str, object]] = {}
```
OliverHeilwagen, line 488: It would be nice to have more descriptive data classes or enum classes to improve the readability.
May also be taken up as a follow up cleanup.

e.g.: 
`provenance`:
```
class PinProvenance(StrEnum):
    ASSERTED = "asserted"
    INCIDENTAL = "incidental"

provenance: dict[str, PinProvenance] = {}
```

`declared_overrides`:
```
@dataclass(frozen=True)
class UncarriedOverride:
    module_name: str
    kind: str
    source_file: str
    reason: str
    consequence: str

    def to_dict(self) -> dict[str, str]:
        return {
            "module": self.module_name,
            "kind": self.kind,
            "file": self.source_file,
            "reason": self.reason,
            "consequence": self.consequence,
        }
declared_overrides: dict[str, UncarriedOverride] = {}
```

`declared_deps`:
```
@dataclass(frozen=True)
class DependencySpec:
    module_name: str
    version: str | None
    dev_dependency: bool
    source_file: str

declared_deps: dict[str, DependencySpec] = {}
```

scripts/known_good/resolved_dependencies.py
```diff
@@ -0,0 +1,1064 @@
...
+                        "consequence": _UNCARRIED_CONSEQUENCE,
+                    },
+                )
+                continue
+            resolved[name] = Module(name=name, hash="", repo="", version=version)
+            provenance[name] = "incidental"
+
+        # The completeness guard: every override ref_int declared that did not become a manifest entry.
+        uncarried = [entry for name, entry in sorted(declared_overrides.items()) if name not in resolved]
+        report = _build_report(resolved, provenance, uncarried, declared_by, _internal_drift(resolved, declared_deps))
```
OliverHeilwagen, line 529: as mentioned above report would also benefit from a dataclass approach e.g.:
```
@dataclass(frozen=True)
class PinReport:
    pin: dict[str, str]
    provenance: PinProvenance
    declared_versions: list[str]
    verdict: Literal["unknown", "by_commit", "agree", "differs"] = "unknown"
    direction: Literal["ref_int_lower", "ref_int_higher"] | None = None

    def to_dict(self) -> dict[str, object]:
        return {
            "pin": self.pin,
            "provenance": self.provenance.value,
            "declared_versions": self.declared_versions,
            "verdict": self.verdict,
            "direction": self.direction,
        }
```

scripts/known_good/resolved_dependencies.py
```diff
@@ -0,0 +1,1064 @@
...
+        elif not module.version:
+            entry["verdict"] = "differs"  # a commit replaced a version request: not comparable
+        elif all(v == module.version for v in wanted):
+            entry["verdict"] = "agree"
+        else:
+            entry["verdict"] = "differs"
+            order = _compare_versions(module.version, _highest(wanted))
+            entry["direction"] = "ref_int_lower" if order < 0 else "ref_int_higher"
+        if entry["verdict"] == "differs":
+            conflicts.append(name)
```
OliverHeilwagen: `conflicts` appears to be a too strong term for the report entry.
I would expect `ref_int_lower` or `ref_int_higher` a typical case. 
Can we stick to `differs` instead, this would also be more self-explanatory.

feature_integration_tests/itf/conftest.py
PiotrKorkus, line 1: this cannot be removed

feature_integration_tests/itf/BUILD
```diff
@@ -69,7 +69,7 @@ py_itf_test(
             "//feature_integration_tests/configs:qemu_bridge_config.json",
             "//images/qnx_x86_64:image",
         ],
-    }) + ["conftest.py"],
```
PiotrKorkus, line 72: its needed

feature_integration_tests/itf/conftest.py
Subramanian-K812: Fixed.

feature_integration_tests/itf/BUILD
Subramanian-K812: Fixed.

scripts/known_good/bazel_version.py
PiotrKorkus, line 1: Bazel version is not dependency to resolve. It is a given from ref_int

patches/logging/005-memory-shared-moved-to-communication.patch
PiotrKorkus, line 1: None of the .bazelrc should be patched. We always take common config from ref_int

patches/persistency/004-safety-level-moved-to-log-rust.patch
PiotrKorkus, line 1: None of the .bazelrc should be patched. We always take common config from ref_int

patches/time/001-memory-shared-moved-to-communication.patch
PiotrKorkus, line 1: None of the .bazelrc should be patched. We always take common config from ref_int

.github/workflows/dr8_test_execution.yml
dcalavrezo-qorix, line 1: every job has `continue-on-error: true`

this means that the workflow can never go red. kind of a no-op as PR check

.github/workflows/dr8_test_execution.yml
```diff
@@ -0,0 +1,284 @@
...
+    branches:
+      - main
+  merge_group:
+    types: [checks_requested]
+  release:
+    types: [created]
+# Do not flood CI with unneeded previous runs in PR
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
+  cancel-in-progress: ${{ github.ref_name != 'main' && !startsWith(github.ref_name, 'release/') }}
```
dcalavrezo-qorix: github.ref_name is the base branch / main so this evaluates to false for PRs  -> old runs never get cancelled.

can you check?

scripts/known_good/resolved_dependencies.py
OliverHeilwagen, line 568: The existing known_good.json makes use of patches (git_override or local_override).
Please also collect the patches so we are capable of injecting them in stage 2.

Related comment is in quality_runner.py
We apply the pachtes locally but the override mechanism is loosing them for dependencies since we use git_override.

OliverHeilwagen, line 669: Please add a warning generation in case the module is using:
- *_override
    - highlight patches? 
- non bazel version resolution dependency
    - http_archive
    - http_file
    - git_repository
    - new_git_repository
    - local_repository

Overrides / Patches used by the module may conflict with the version imposed by reference integration.
Ideally we should do such a scan as a initial sanity check if the versions of known_good.json are updated.

If a selected module contain such elements it may cause build / test conflicts, being aware early could reduce time for issue identification.

OliverHeilwagen, line 896: For simplified testability i would recommend to extract the regular expression functionalities into a separate parser class. Improvement can be done as a follow up PR.

.github/workflows/dr8_test_execution.yml
```diff
@@ -0,0 +1,284 @@
...
+        if: always()
+        run: |
+          mkdir -p artifacts/stage1-resolved-deps
+          # Merge the resolved registry versions with ref_int's own override directives into the
+          # single Stage 1 -> Stage 2 handoff manifest. The script stores the graph alongside it,
+          # which Stage 2 needs to pin each module's full transitive closure.
+          bazel mod graph --output=json --lockfile_mode=error > resolved_graph.json
+          python scripts/known_good/resolved_dependencies.py \
+            --mod-graph resolved_graph.json \
+            --export artifacts/stage1-resolved-deps/resolved_versions.json
```
OliverHeilwagen: Switch to bazel command is now also possible this way we would not be affected by a python change of the underlying image used by the github runner.

scripts/quality_runners.py
```diff
@@ -302,46 +644,181 @@ def parse_arguments() -> argparse.Namespace:
...
+                "Stage-1 resolved set, which known_good.json cannot supply."
+            )
+        # Before overwrite(): a patch may add a bazel_dep, and overwrite() reads the declared
+        # deps to decide which closure members still need one.
+        module_under_test = known.modules["target_sw"].get(args.modules_to_test[0])
+        if module_under_test is None:
+            raise SystemExit(f"{args.modules_to_test[0]} is not a target_sw module in {args.known_good_path}")
+        print_centered(f"QR: Applying {module_under_test.name}'s own bazel_patches to {workspace}")
+        try:
+            apply_module_patches(module_under_test, workspace.resolve(), REF_INT_ROOT, log=print_centered)
```
OliverHeilwagen, line 794: patches are applied locally but the resolved.overwrite() patches are using single_version_override or git_override which leads to missing patches for dependencies in the stage 2 module run.

Ideally we can transport a patch of a dependency into the overwrite patching functionality.

.github/workflows/dr8_test_execution.yml
```diff
@@ -0,0 +1,284 @@
...
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+name: DR-008 Test Execution (Stage 1 & 2)
+permissions:
+  contents: write
+  pull-requests: write
+on:
+  pull_request_target: # Allow forks to access secrets
```
OliverHeilwagen: This should only be required once we extend to QNX build and test in this workflow.
Since we use `--config=linux-x86_64` shouldn't the following be sufficient?
Then we can also check the run in this PR.

```suggestion
  pull_request:
```

scripts/known_good/bazel_version.py
```diff
@@ -0,0 +1,59 @@
...
+# rather than guessed at.
+_RELEASE_RE = re.compile(r"\d+(?:\.\d+)*")
+
+
+def _order(version: str) -> tuple[int, ...] | None:
+    """The comparable form of a release string, or None if it is not a plain dotted release."""
+    return tuple(int(part) for part in version.split(".")) if _RELEASE_RE.fullmatch(version) else None
+
+
+def resolve_stage2_bazel_version(ref_int_version: str, module_version: str | None) -> str:
```
OliverHeilwagen: Why don't we demand a .bazelversion to be present under root of the module?
This way we could simplify the call and just give the root path of ref_int and module the module.
We could then also have the bazel version patching logic in this python file.

.github/workflows/dr8_test_execution.yml
```diff
@@ -0,0 +1,284 @@
...
+      - name: Checkout repository
+        if: ${{ github.event_name != 'pull_request_target' }}
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - name: Execute Feature Integration Tests
+        run: |
+          bazel test --lockfile_mode=error --config=linux-x86_64 //feature_integration_tests/test_cases:fit
+      - name: Export resolved dependency manifest
+        if: always()
+        run: |
+          mkdir -p artifacts/stage1-resolved-deps
```
OliverHeilwagen, line 60: Please provide all the steps also in a README for local replication of the workflow.

scripts/known_good/verify_stage2_resolution.py
```diff
@@ -0,0 +1,134 @@
...
+        injected = injected_override_names(args.module_bazel.read_text(encoding="utf-8"))
+
+    for path in (args.mod_graph, args.resolved):
+        if not path.is_file():
+            print(f"::error::verify_stage2_resolution: missing {path}")
+            return 1
+
+    actual = module_graph_versions(args.mod_graph)
+    resolved = json.loads(args.resolved.read_text(encoding="utf-8"))
+    resolved = resolved.get("modules", resolved)
```
OliverHeilwagen, line 129: since we base the check upon the bazel mod graph of the module and the resolved_version.json which both do not contain information about patches demanded by reference integration, the check is incomplete.

OliverHeilwagen (changes requested):
review focus ./scripts/known_good, excluding tests.
Transitive dependency are now correctly pinned 👍 .

Patches from reference integration are locally applied but patching in Stage 2 uses single_version_override and git_override without patches, which leads to missing patches for module dependencies, which were present in Stage 1.

.github/workflows/dr8_test_execution.yml
MaximilianSoerenPollak: Agreed, pull_request_target should not be used anywhere except where strictly needed.

.github/workflows/dr8_test_execution.yml
Subramanian-K812: Removed from all four jobs — the workflow now goes red when Stage 1 or Stage 2 fails.

scripts/known_good/resolved_dependencies.py
Subramanian-K812: Renamed to differs, matching the per-pin verdicts.

scripts/known_good/resolved_dependencies.py
Subramanian-K812: Added module_resolution_hazards() — warns on a module's own *_override and on http_archive-style fetches.

.github/workflows/dr8_test_execution.yml
Subramanian-K812: Switched — PR now exercises its own workflow.

.github/workflows/dr8_test_execution.yml
Subramanian-K812: Added ci/README.md covering every step.

scripts/known_good/bazel_version.py
Subramanian-K812: bazel_version.py is removed; that logic now lives in quality_runners.py and always imposes ref_int's version.

scripts/known_good/resolved_dependencies.py
Subramanian-K812: Agreed — taking these as a follow-up PR, since this one already reworks the same file for patch transport.

scripts/known_good/resolved_dependencies.py
Subramanian-K812: Agreed — follow-up PR, keeping this one scoped to the patch-transport fix.

scripts/known_good/bazel_version.py
Subramanian-K812: Agreed — bazel_version.py is deleted; Stage 2 now just stamps ref_int's .bazelversion onto the module checkout

patches/logging/005-memory-shared-moved-to-communication.patch
Subramanian-K812: Removed — the patch now redirects source labels only.

patches/persistency/004-safety-level-moved-to-log-rust.patch
Subramanian-K812: Removed — this patch is deleted entirely.

patches/time/001-memory-shared-moved-to-communication.patch
Subramanian-K812: Removed — the patch now redirects source labels only.

.github/workflows/dr8_test_execution.yml
Subramanian-K812: Fixed — cancel-in-progress no longer reads ref_name.

scripts/known_good/resolved_dependencies.py
Subramanian-K812: Done — _parse_override_file now collects patches/patch_strip

scripts/known_good/verify_stage2_resolution.py
Subramanian-K812: Complete now — the manifest carries patches, and missing_patches() fails the check when an injected dep lost them.

.github/workflows/dr8_test_execution.yml
Subramanian-K812: Done — all four scripts now run as bazel run py_binary targets

scripts/known_good/resolved_dependencies.py
Subramanian-K812: Agreed — taking this as a follow-up PR

scripts/quality_runners.py
Subramanian-K812: Fixed — Stage 1 now ships the patch bytes alongside the resolved dependency set, and overwrite() re-hosts them inside the checkout as //ref_int_patches:..., so injected dependencies build patched instead of losing their patches.

OliverHeilwagen:
@Subramanian-K812 Thanks a lot for the quick rework 👍 . Unfortunately I am not at work till 09.09., once I am back I can continue the re-review.

.github/workflows/dr8_test_execution.yml
```diff
@@ -0,0 +1,251 @@
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Publish consolidated quality report
+        if: always()
+        run: |
+          bazel run --ui_event_filters=-info,-stdout --noshow_progress \
+            //scripts:aggregate_quality_report -- \
+            --stage1-result "${{ needs.stage1_integration.result }}" \
+            --stage2-result "${{ needs.stage2_module_validation.result }}" \
+            --stage2-dir "_stage2_reports/" \
+            >> "$GITHUB_STEP_SUMMARY"
```
MaximilianSoerenPollak, line 251: Is there an output lenght limit for the step summary? 
If there is, will this hit it or is there no risk of that?

patches/time/001-memory-shared-moved-to-communication.patch
```diff
@@ -3,6 +3,8 @@ b90e0695 ("Migrate score/memory/shared to score_communication"). score_time HEAD
 still points at the old @score_baselibs labels, so redirect them and declare the
 score_communication dependency they now require.
 
+Source labels only; the module's .bazelrc has the same stale labels but ref_int no longer patches it.
```
MaximilianSoerenPollak, line 6: But if it's stale labels should they not be removed?

scripts/known_good/BUILD
```diff
@@ -22,6 +22,9 @@ py_library(
         ["**/*.py"],
         exclude = ["tests/**"],
     ),
+    # scripts/ on sys.path, so "from known_good...." resolves under 'bazel run' as it does for
+    # "python3 scripts/...". Propagates to every consumer of this library.
+    imports = [".."],
```
MaximilianSoerenPollak, line 27: Are you sure this does not break anything else? 

Importing `..` can have unexpected consequences in consumers of this py_library.

scripts/known_good/list_modules.py
```diff
@@ -0,0 +1,88 @@
...
+import argparse
+import json
+import re
+import sys
+from pathlib import Path
+
+
+def repo_slug(repo_url: str) -> str:
+    """Derive the 'owner/name' slug actions/checkout expects from a git URL."""
+    match = re.search(r"[:/]([^/:]+/[^/:]+?)(?:\.git)?/?$", repo_url or "")
```
MaximilianSoerenPollak, line 39: Very sure this regex does more and less than what it is suppose to do (as all regex). 
But I guess it's okay for now?

scripts/known_good/list_modules.py
```diff
@@ -0,0 +1,88 @@
...
+
+_HERE = Path(__file__).resolve().parent
+try:
+    from known_good.models.known_good import load_known_good
+    from known_good.resolved_dependencies import repo_root, workspace_path
+except ImportError:
+    if str(_HERE) not in sys.path:
+        sys.path.insert(0, str(_HERE))
+    from models.known_good import load_known_good  # noqa: E402
+    from resolved_dependencies import repo_root, workspace_path  # noqa: E402
```
MaximilianSoerenPollak, line 51: Why is this needed again?

scripts/known_good/update_module_from_known_good.py
```diff
@@ -47,64 +57,10 @@ def generate_git_override_blocks(modules: List[Module], repo_commit_dict: Dict[s
...
-                f'bazel_dep(name = "{module.name}")\n'
-                "git_override(\n"
-                f'    module_name = "{module.name}",\n'
-                f'    commit = "{commit}",\n'
-                f"{patch_strip_line}"
-                f"{patches_lines}"
-                f'    remote = "{module.repo}",\n'
-                ")\n"
-            )
-        blocks.append(block)
```
MaximilianSoerenPollak, line 107: This whole logic not needed anymore, or moved somewhere else?

scripts/known_good/verify_stage2_resolution.py
```diff
@@ -0,0 +1,187 @@
...
+except ImportError:
+    if str(_HERE) not in sys.path:
+        sys.path.insert(0, str(_HERE))
+    from module_patches import patch_relpath  # noqa: E402
+    from resolved_dependencies import (  # noqa: E402
+        INJECTED_PATCHES_PKG,
+        _collect_resolved_versions,
+        injected_override_names,
+        workspace_path,
+    )
```
MaximilianSoerenPollak, line 59: I'm sure there is a better way than to this try / except. 

What are you trying to achieve with this again?

scripts/known_good/verify_stage2_resolution.py
```diff
@@ -0,0 +1,187 @@
+#!/usr/bin/env python3
```
MaximilianSoerenPollak, line 1: If this file runs in CI, be careful that we already had the issue (in docs-as-code) that we ran into the upper limit of a pipe buffer in the CI output (64kb) due to folding it with long prints. 
Just an FYI

MaximilianSoerenPollak (commented):
Some questions from my side. 
But overall I think it does what was asked according to the DR.

PiotrKorkus (commented):
- nested bazel invokes in `ferrocene_report.sh` coming from the toolchain do not carry `--noworkspace_rc` which makes rust coverage fail
- huge differences in number of executed tests. e.g. communication has 6500 test cases under `//score/...` but here we run only 3800. Why do we have differences when we should execute from the score of tested repo just with common config and deps?
- how about appending single line with import of common bazelrc to the one in tested module and using the config from it? Would that work or?

scripts/known_good/update_module_from_known_good.py
Subramanian-K812: Moved, not removed — now imported from resolved_dependencies.generate_override_directive so Stage 1 and Stage 2 can't emit different git_override syntax.

scripts/known_good/verify_stage2_resolution.py
Subramanian-K812: Thanks for the heads-up, this file's output isn't piped into anything in CI — it runs as a plain bazel run step straight to the job log, so the pipe-buffer deadlock docs-as-code hit shouldn't be an issue here.

.github/workflows/dr8_test_execution.yml
Subramanian-K812: Checked against GitHub's actual limit (1 MiB per step, upload fails with an annotation if exceeded) — a full 8-module report from the real script is ~4.3 KB, so this should not reach the cap.

scripts/known_good/BUILD
Subramanian-K812: Yes agreed — better to scope or drop imports than leave it shared.

patches/time/001-memory-shared-moved-to-communication.patch
Subramanian-K812: We could, but that would require patching the module’s .bazelrc, which has already been ruled out for this PR.

scripts/known_good/verify_stage2_resolution.py
Subramanian-K812: Agreed — the current try/except isn’t ideal; it’s a leftover from the same pattern used in files that are still invoked directly, which don’t have py_binary targets, so those workflows call the .py files directly instead of through bazel run.

scripts/known_good/list_modules.py
Subramanian-K812: This to be improved  — switching to urllib.parse for the path url in known_good.json.

scripts/known_good/list_modules.py
Subramanian-K812: Agreed — same as noted below, the current try/except isn’t ideal and will be improved in a follow‑up.

scripts/known_good/resolved_dependencies.py
OliverHeilwagen: Please resolve comment.

scripts/known_good/resolved_dependencies.py
OliverHeilwagen: Please resolve comment.

scripts/known_good/resolved_dependencies.py
OliverHeilwagen: Please resolve comment.

OliverHeilwagen: Please also take over the bazel_patches for a version pin.

```
        if module.bazel_patches:
            entry["pin"]["bazel_patches"] = module.bazel_patches
```

OliverHeilwagen: Please keep the description free of examples

scripts/known_good/resolved_dependencies.py
```diff
@@ -647,6 +728,17 @@ def overwrite(
         # module's own override for every dep we inject so ref_int's resolved version wins.
         original = _strip_existing_overrides(original, injected_names)
 
+        # After the module's competing overrides are stripped and before ref_int's are appended,
+        # what is left is the module's own resolution that survives injection.
+        hazards = module_resolution_hazards(original)
```
OliverHeilwagen: Thanks for adding this check.

Out of scope for this PR.
In general i would wish for such a sanity check in `stage 0`, especially if we want to increase the frequency updating reference_integration. We should be informed early about *_override, patches or other 'anti patterns' within modules, which may cause issues during integration.
Hopefully this can reduce the time to analyse issues when updating `known_good.json`.

scripts/known_good/resolved_dependencies.py
```diff
@@ -612,16 +682,19 @@ def overwrite(
...
             if module is None:
                 unresolved.append(name)
                 continue
-            # Strip bazel_patches: they reference //patches/... labels in ref_int's
-            # workspace which do not exist inside another module's checkout.
-            module = _replace(module, bazel_patches=None)
+            labels = self._stage_patches(module, workspace, patch_source)
+            if module.bazel_patches and not labels:
+                unpatched.append(f"{name} ({len(module.bazel_patches)})")
+            module = _replace(module, bazel_patches=labels or None)
```
OliverHeilwagen: Out of scope for this PR:
For improved readability creating the correct patch path in the override directive should be a sub method or within generate_override_directive.

However such readability refactoring are not mandatory for the initial introduction of the stage1 + stage2 approach.

scripts/known_good/resolved_dependencies.py
```diff
@@ -562,6 +628,7 @@ def overwrite(
         graph: DependencyGraph,
         *,
         module_under_test: str | None = None,
+        patch_source: Path | None = None,
```
OliverHeilwagen: Stage 2 bazel_dep patching is now from my point of view complete and covers all cases.

What out fo scope for this PR from my point of view is the toolchain related patches.
In stage 2 the toolchain configuration of the module is kept (e.g. QNX sdp version), which is incorrect from the big picture.

Since these updates are less frequent i don't see it as a high priority for now but the topic should be planned in the backlog.

scripts/known_good/list_modules.py
```diff
@@ -0,0 +1,88 @@
...
+
+def main() -> None:
+    parser = argparse.ArgumentParser(description="List known_good.json modules of a group as JSON (for CI matrices).")
+    parser.add_argument(
+        "--known-good-path",
+        type=Path,
+        default=repo_root() / "known_good.json",
+        help="Path to known_good.json (default: repo-root known_good.json).",
+    )
+    parser.add_argument("--group", default="target_sw", help="Module group to list (default: target_sw).")
```
OliverHeilwagen: I would recommend to add `choices=("target_sw", "tooling"),`

.github/workflows/internal_tests.yml
```diff
@@ -21,4 +21,6 @@ jobs:
   internal_tests:
     uses: eclipse-score/cicd-workflows/.github/workflows/tests.yml@main
     with:
-      bazel-target: "test //scripts/tooling:tooling_tests //scripts/known_good:known_good_tests"
+      # Bundles tooling_tests with known_good_tests and quality_scripts_tests; previously only
```
OliverHeilwagen: General - out of scope for this PR.

Since most of the tooling of ref_int will have a high iso26262 tcl level (2-3).
We should at some point start measures for tool qualification, my expectation would be brief tool requirements, integration tests, unit tests and coverage for the python scripts.
Is this already considered on the roadmap?

scripts/known_good/verify_stage2_resolution.py
```diff
@@ -0,0 +1,187 @@
...
+    agreed = 0
+    by_commit = 0
+
+    for name, entry in sorted(resolved.items()):
+        if name == args.module:
+            continue
+        want = entry.get("version")
+        if want is None:
+            by_commit += 1  # git_override'd: pinned by commit, not comparable by version
+            continue
```
OliverHeilwagen: Since MODULE.bazel is source of truth for git_overrides can we add this as a data input to complete the check?
It is a little bit paranoid but would be nice round trip verification.

scripts/quality_runners.py
```diff
@@ -302,45 +711,208 @@ def parse_arguments() -> argparse.Namespace:
...
+            print_centered(
+                f"QR: Layering ref_int config {STAGE2_RC} over {args.modules_to_test[0]}'s own "
+                ".bazelrc (registers a libclang toolchain ref_int cannot supply)"
+            )
+        elif dedicated_rc is not None:
+            print_centered(
+                f"QR: Using ref_int config {STAGE2_RC} plus {dedicated_rc}; the module's .bazelrc is not read"
+            )
+        else:
+            print_centered(f"QR: Using ref_int config {STAGE2_RC} only; the module's .bazelrc is not read")
```
OliverHeilwagen: Optional for this PR:
It would be nice to refactor the existing functionality into more readable steps for the `main()`.
- select_bazelrc
- apply_patches
- inject_resolved_dependencies
- update_module_lock (currently remove)
- pin_bazel_version
- ...

e.g. wrapper class that can also store data and does not need to transport it in main to the next method.

scripts/quality_runners.py
```diff
@@ -34,115 +49,492 @@ def print_centered(message: str, width: int = 120, fillchar: str = "-") -> None:
...
+
+    Failures are reported and swallowed -- evidence capture must never fail a passing gate. A
+    non-zero exit is expected for some modules, because ``mod graph`` evaluates every extension and
+    so inherits the over-reach that disqualified ``mod deps`` (see :func:`run_resolution_gate`); it
+    still prints the complete graph, which the DR-008 verification step reads. So the dump is kept
+    whenever it parses and the exit code is reported rather than obeyed.
+    """
+    graph_path = workspace.resolve() / MODULE_GRAPH_NAME
+    call = ["bazel"] + (startup or []) + ["mod", "graph", "--output=json", "--lockfile_mode=update"]
+    result = run_command(call, cwd=str(workspace))
```
OliverHeilwagen: Please dump directly into file or is there a reason to see the full module graph in the log?

scripts/quality_runners.py
```diff
@@ -34,115 +49,492 @@ def print_centered(message: str, width: int = 120, fillchar: str = "-") -> None:
...
+    try:
+        injected = injected_override_names(module_bazel.read_text())
+    except OSError:
+        return "ref_int harness defect", []
+    # Bazel spells an injected module's repo as @@<name>+ (canonical) or @<name>// (apparent).
+    named = sorted(n for n in injected if f"@@{n}+" in output or f"@{n}//" in output)
+    return ("integration conflict", named) if named else ("ref_int harness defect", [])
+
+
+def run_unit_test_with_coverage(
```
OliverHeilwagen: Out of scope for this PR.
Would be nice to refactor the individual quality checks into their own classes | files, so we can import them in `quality_runners` and invoke them from `main()`.

ci/stage2/module.bazelrc
```diff
@@ -0,0 +1,98 @@
...
+# score_logging has no Rust instrumentation config, so this is its only source.
+build:ferrocene-coverage --@rules_rust//rust/settings:extra_rustc_flag=-Cinstrument-coverage
+build:ferrocene-coverage --@rules_rust//rust/settings:extra_rustc_flag=-Clink-dead-code
+build:ferrocene-coverage --@rules_rust//rust/settings:extra_rustc_flag=-Ccodegen-units=1
+build:ferrocene-coverage --@rules_rust//rust/settings:extra_rustc_flag=-Cdebuginfo=2
+build:ferrocene-coverage --@rules_rust//rust/settings:extra_exec_rustc_flag=-Cinstrument-coverage
+build:ferrocene-coverage --@rules_rust//rust/settings:extra_exec_rustc_flag=-Clink-dead-code
+build:ferrocene-coverage --@rules_rust//rust/settings:extra_exec_rustc_flag=-Ccodegen-units=1
+build:ferrocene-coverage --@rules_rust//rust/settings:extra_exec_rustc_flag=-Cdebuginfo=2
+test:ferrocene-coverage --run_under=@score_tooling//coverage:llvm_profile_wrapper
```
OliverHeilwagen: I guess we should already change this to `score_coverage`.
https://github.com/eclipse-score/score/issues/3258

OliverHeilwagen (approved):
Approve with suggestions.

Tests for the scripts were not in scope of this review.

From my point this is a nice initial version, which we can further be refactored to simplify the test complexity for the scripts. (ISO26262 tool qualification depending on the classification level).

Toolchain configuration overwrites for stage 2 are not be expected from my side for this PR, should however be planned in the backlog.

Current failing stage 2 steps need to be resolved, but are related to integration details (e.g. upgrade of coverage tooling score_tooling -> score_coverage) and not caused by general points in modified | added scripts.

PR #282, opened 2026-07-09
Remove unnecessary stuff

LittleHuba:


pawelrutkaq:
please dont merge

LittleHuba:
There is no intention to merge this from my side. This is to highlight the wrong testing strategy that forces all these patches that users should never ever do.

pawelrutkaq:
The DR8 will be implemented for 0.9 ;)

PR #299, opened 2026-07-27
feat(deps): add score_config_management to reference integration

Chahult:
- Added score_config_management module (commit ) to known_good.json under target_sw with code_root_path //score/config_management/...

- Regenerate bazel_common/score_modules_target_sw.MODULE.bazel


Chahult:
Hi @pawelrutkaq @PiotrKorkus @nradakovic @antonkri @AlexanderLanin , could someone please unlock this workflow and review our changes? Thanks!

LittleHuba (dismissed):

Chahult:
Updated the baselibs to the latest version as builds are breaking.

LittleHuba:
> Updated the baselibs to the latest version as builds are breaking.

Please bump to the version pulled in by communication. That gives you a good chance to have a successful build.

Chahult:
Hi @pawelrutkaq  @PiotrKorkus , could you please review and approve these changes? The aim of this PR is to add `score_config_management` to the reference integration.

michaelsaborov (dismissed):

known_good.json
```diff
@@ -135,6 +135,29 @@
                         "cpp"
                     ]
                 }
+            },
+            "score_config_management": {
+                "repo": "https://github.com/eclipse-score/config_management.git",
+                "hash": "b22ca9bd222b99985be4339187c639d6862051a0",
+                "bazel_patches": [
+                    "patches/config_management/001-fix-communication-api-and-remove-targets.patch"
```
pawelrutkaq: Since this looks like an arbitrary patch, someone from com shall ack it.

known_good.json
```diff
@@ -135,6 +135,29 @@
...
+                    "code_root_path": "//score/config_management/...",
+                    "extra_test_config": [
+                        "@score_config_management//score/config_management/config_daemon/code:score_variant=true"
+                    ],
+                    "exclude_test_targets": [
+                        "//score/config_management/config_daemon/code/factory/details:unit_test_mw_com",
+                        "//score/config_management/config_daemon/code/services/details/mw_com:unit_test",
+                        "//score/config_management/config_provider/code/config_provider/factory:unit_tests_mw_com",
+                        "//score/config_management/config_provider/code/proxies/details:unit_test_mw",
+                        "//score/config_management/dependability/..."
```
pawelrutkaq, line 155: why we need those exclusions ?

pawelrutkaq (changes requested):
Two points
- please deploy the example (from your repo) in showcases as other modules does
- please at least setup correctly module in bazel registry (no need of release ) -> already done

pawelrutkaq (changes requested):
If comm has already this change in, I would suggest to update com as well here.

known_good.json
Chahult: 1.mw_com tests: require a running LoLa/SOME/IP runtime — the tests themselves are tagged "exclusive" because "Sandboxing is not properly working with LoLa message passing 2.0". Same reason score_communication excludes //score/mw/com/impl:unit_test_runtime_single_exec and //score/mw/com/impl:runtime_test.

2.dependability/...: these are component/dependable_element safety-metadata targets from score_tooling, not executable tests. Including //... without filtering them would cause build failures.

known_good.json
Chahult: The current name 001-fix-communication-api-and-remove-targets.patch is misleading — the patch doesn't fix the communication library, it adapts config_management to the new ProxyFuture API and removes internal/ARA-specific targets. Let me rename it

pawelrutkaq: Please do not integrate new module with patches, please fix it upstream

Chahult: The OptionalProxyData → ProxyFuture part of the patch is a temporary mismatch caused by an in-progress mw::com migration: the internal main repo still uses the old OptionalProxyData API, while the OSS communication repo has already migrated to ProxyFuture. Once that migration completes and syncs upstream, config_management's main repo will use ProxyFuture too, and this part of the patch will no longer be needed — no upstream fix is possible for this piece yet.

pawelrutkaq:
I see that there are some errros in CFM example (especially from CFG daemon):
```
63ns Starting example 'Config Management Daemon Example'
6.54µs App 1: starting /showcases/bin/config_daemon
App 1: spawned command cd "/showcases/data/config_management" && "/showcases/bin/config_daemon" "-s" "/showcases/data/config_management/etc/mw_com_config.json"
mw::log initialization error: Error No logging configuration files could be found. occurred with context information: Failed to load configuration files. Fallback to console logging.
2026/08/13 06:22:57.2177293 4153009148 000 ECU1 NONE lola log warn verbose 1 <maxSamples> property for event is DEPRECATED! use <numberOfSampleSlots> property for event
2026/08/13 06:22:57.2177293 4153009156 000 ECU1 NONE LM log error verbose 3 [Lifecycle client] FD  111  is invalid for kRunning report
2026/08/13 06:22:57.2177293 4153009156 000 ECU1 NONE App log error verbose 3 ConfigDaemon:: Run Failed to create InitialQualifierStateSender callback
2026/08/13 06:22:57.2177293 4153009156 000 ECU1 NONE DFLT log error verbose 1 Error occurred during Run
```

NicolasFussberger:
> I see that there are some errros in CFM example (especially from CFG daemon):
> 
> ```
> 63ns Starting example 'Config Management Daemon Example'
> 6.54µs App 1: starting /showcases/bin/config_daemon
> App 1: spawned command cd "/showcases/data/config_management" && "/showcases/bin/config_daemon" "-s" "/showcases/data/config_management/etc/mw_com_config.json"
> mw::log initialization error: Error No logging configuration files could be found. occurred with context information: Failed to load configuration files. Fallback to console logging.
> 2026/08/13 06:22:57.2177293 4153009148 000 ECU1 NONE lola log warn verbose 1 <maxSamples> property for event is DEPRECATED! use <numberOfSampleSlots> property for event
> 2026/08/13 06:22:57.2177293 4153009156 000 ECU1 NONE LM log error verbose 3 [Lifecycle client] FD  111  is invalid for kRunning report
> 2026/08/13 06:22:57.2177293 4153009156 000 ECU1 NONE App log error verbose 3 ConfigDaemon:: Run Failed to create InitialQualifierStateSender callback
> 2026/08/13 06:22:57.2177293 4153009156 000 ECU1 NONE DFLT log error verbose 1 Error occurred during Run
> ```

I assume the error regarding lifecycle is that the config manager uses the lifecycle API to report running state, but it is not started by Launch Manager so the IPC channel is not actually available.
I think this has no other consequences for the demo despite the error log.

Chahult:
**Error happens in config_management:**
score::mw::service (the generic service-registration framework in communication) has no real implementation yet — only a stub, added without verifying it actually works with ConfigDaemon.
Even with that fixed, the "demo" only proves ConfigDaemon can call OfferService() successfully and stay alive — it's an in-process stub, not the real LoLa/mw::com IPC backend (that migration is still ongoing in SPP).
**Root conclusion:** with only a stub mw::service, we can show the daemon starts and doesn't crash, but we cannot demonstrate any real service behavior, because there's no client and no real backend to show data flowing.

**Proposed path forward:**
Preferred: merge score_config_management into the reference_integration PR without a demo example — avoids implying functionality that doesn't exist yet.
If a demo is mandatory: provide a stub demo with no real service implementation — i.e., just start ConfigDaemon, confirm it comes up and stays alive without crashing, purely to unblock the PR from pending state. This would not demonstrate the service actually being consumed.
We can revisit a real, meaningful demo once mw::service has a proper implementation and the SPP mw::com migration lands.

pawelrutkaq:
Please make PR green, in the example description, write `non functional yet` and we are good. Keep in mind that the patch You add shall be removed until 0.9, which is 07.09.2026

Chahult:
> Please make PR green, in the example description, write `non functional yet` and we are good. Keep in mind that the patch You add shall be removed until 0.9, which is 07.09.2026

Sure, currently we are working on that part where we will remove this patch and add a proper demo example.

pawelrutkaq (approved):

PR #300, opened 2026-07-27
test(fit): add score_time clock integration test

clanghans:
## What

Adds a feature integration test (FIT) for the `score_time` clock library.
`score_time` is already pinned in `known_good.json` (and the generated
`bazel_common/score_modules_target_sw.MODULE.bazel`), but so far it had no
integration test coverage. This adds one following the existing FIT pattern
(C++ scenario binary + pytest driver).

## Scope

Three C++ scenarios exercise the public `Clock<Tag>` API and self-validate,
throwing (process exit 101) on failure:

- `time.system_clock_now` — `SystemClock::Now()` tracks the host system clock
  within a generous tolerance (catches an epoch-zero stub / unlinked backend).
- `time.steady_clock_now` — two `SteadyClock::Now()` readings are monotonic.
- `time.high_res_steady_clock_now` — `HighResSteadyClock::Now()` is non-zero and
  monotonic.

Each scenario emits its reading as a structured JSON log line. The pytest FIT
case (`tests/time/test_clock_now.py`) drives each scenario and asserts the exit
code plus the emitted reading. score_time is C++ only, so the tests are
parametrized with `version=["cpp"]`.

Deferred (not in this PR): VehicleTime (requires explicit `Init()`), and the
time_daemon / time_slave / ts_client processes (better suited to ITF/QEMU tests).

## Files

- `feature_integration_tests/test_scenarios/cpp/src/scenarios/time/` — new scenarios + group `mod.cpp`
- `feature_integration_tests/test_scenarios/cpp/src/internals/time/clock_log.h` — small JSON log helper
- `feature_integration_tests/test_scenarios/cpp/src/scenarios/mod.cpp` — register the `time` group
- `feature_integration_tests/test_scenarios/cpp/BUILD` — add the three `@score_time` clock deps
- `feature_integration_tests/test_cases/tests/time/test_clock_now.py` — pytest FIT case

## Verification

- `bazel build --config=linux-x86_64 //feature_integration_tests/test_scenarios/cpp:cpp_test_scenarios` — builds with the score_time deps.
- Ran each scenario binary directly — exit 0 with valid readings.
- `bazel test --config=linux-x86_64 //feature_integration_tests/test_cases:fit_cpp` — all tests pass (6 new time tests + existing suite).
- Python (ruff) and Starlark (buildifier) format checks pass.

### DIFF
--- a/feature_integration_tests/test_cases/tests/time/test_clock_now.py
+++ b/feature_integration_tests/test_cases/tests/time/test_clock_now.py
@@ -1,0 +1,100 @@
+# *******************************************************************************
+# Copyright (c) 2026 Contributors to the Eclipse Foundation
+#
+# See the NOTICE file(s) distributed with this work for additional
+# information regarding copyright ownership.
+#
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+
+from typing import Any
+
+import pytest
+from fit_scenario import FitScenario, ResultCode
+from testing_utils import BazelTools, BuildTools, LogContainer, ScenarioResult
+
+# score_time is C++ only; there is no Rust variant of the clock library.
+pytestmark = pytest.mark.parametrize("version", ["cpp"], scope="class")
+
+
+class ClockScenario(FitScenario):
+    """Common base for score_time clock scenarios (no scenario input required)."""
+
+    @pytest.fixture(scope="class")
+    def build_tools(self, version: str) -> BuildTools:
+        # Consume the parametrized `version` and select the C++ scenario binary.
+        return BazelTools(option_prefix=version)
+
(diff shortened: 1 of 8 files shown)


feature_integration_tests/test_cases/tests/time/test_clock_now.py
```diff
@@ -0,0 +1,100 @@
...
+
+import pytest
+from fit_scenario import FitScenario, ResultCode
+from testing_utils import BazelTools, BuildTools, LogContainer, ScenarioResult
+
+# score_time is C++ only; there is no Rust variant of the clock library.
+pytestmark = pytest.mark.parametrize("version", ["cpp"], scope="class")
+
+
+class ClockScenario(FitScenario):
```
PiotrKorkus, line 24: redundant? should be possible to use FitScenario directly

PiotrKorkus (changes requested):
good for basic smoke tests

did you consider any feature reqs? could you add the mapping?

I would suggest adding a scenario where we do a couple of ticks in a loop, assert all are not decreasing, then sleep for some known duration and compare again next tick.

